PR #18, opened 2015-08-12
Installation, SysV, Systemd Support

akutz:
This patch adds all sorts of wild and wonderful things to the REX-Ray
binary. For starters, installation. To install REX-Ray on a supported
Linux distribution, all that is required now is to download the binary
and execute:

    sudo ./rexray install

What does that do? In short the above command will determine if the
Linux distribution uses systemctl, update-rc.d, or chkconfig to manage
system services. After that the following steps occur:

  1. The path /opt/rexray is created and chowned to root:root with
     permissions set to 0755.

  2. The binary is copied to /opt/rexray/rexray and chowned to
     root:root with permissions set to 4755. This is important, because
     this means that any non-privileged user can execute the rexray
     binary as root without requiring sudo privileges. For more
     information on this feature, please read about the Linux kernel's
     super-user ID (SUID) bit at
     http://www.tldp.org/HOWTO/Security-HOWTO/file-security.html.

     Because the REX-Ray binary can now be executed with root
     privileges by non-root users, the binary can be used by non-root
     users to easily attach and mount external storage.

  3. The directory /etc/rexray is created and chowned to root:root.

The next steps depends on the type of Linux distribution. However, it's
important to know that the new version of the REX-Ray binary now
supports managing its own PID (at /var/run/rexray.pid) when run as a
daemon as well as supports the standard SysV control commands such as
"start," "stop," "status," and "restart."

For SysV Linux distributions that use chkconfig or update-rc.d, a
symlink of the REX-Ray binary is created in /etc/init.d/ and then
either "chkconfig rexray on" or "update-rc.d rexray defaults" is
executed.

Modern Linux distributions have moved to SystemD for controlling
services. If the "systemctl" command is detected when installing
REX-Ray then a unit file is written to
/etc/systemd/system/rexray.service with the following contents:

    [Unit]
    Description=rexray
    Before=docker.service

    [Service]
    EnvironmentFile=/etc/rexray/rexray.env
    ExecStart=/opt/rexray/rexray --daemon
    ExecReload=/bin/kill -HUP $MAINPID
    KillMode=process
    Restart=always
    StartLimitInterval=0

    [Install]
    WantedBy=docker.service

Additionally, an environment file (and parent directory) is created at
/etc/rexray/rexray.env with the contents:

    REXRAY_HOME=/opt/rexray

The REX-Ray service is not started immediately upon installation. The
install command completes by informing the users that they should visit
http://github.com/emccode/rexray for information on how to configure
REX-Ray's storage drivers. The text to the users also explains how to
start the REX-Ray daemon once it's configured using the service command
particular to the Linux distribution.


akutz:
I am closing this pull request to reopen with the latest changes as this PR is too far from what was originally offered.

Issue #2, opened 2015-06-30
Daemon mode initializes when no storage adapters are present


clintonskitson:
The daemon should not initialize if there are no storage drivers present to be used.  This way when running rexray as a service, it will retry starts until a storage driver can be initialized successfuly.

clintonskitson:
See PR https://github.com/emccode/rexray/pull/3

PR #19, opened 2015-08-13
added volume options on creation for docker daemon

clintonskitson:
This PR accommodates upcoming changes in the Docker Volume API.  The change is that a map[string]string is now passed for creation options.

This enables commands like the following directly from Docker.
docker volume create --driver=rexray --opt=size=16 --opt=IOPS=100 --opt=volumetype=ssd

https://github.com/docker/docker/pull/14242/files

### DIFF
--- a/daemon/module/docker/remotevolumedriver/remvoldriver.go
+++ b/daemon/module/docker/remotevolumedriver/remvoldriver.go
@@ -112,8 +112,9 @@ var (
 )
 
 type pluginRequest struct {
-	Name       string `json:"Name,ommitempty"`
-	InstanceID string `json:"Instanceid,ommitempty"`
+	Name       string            `json:"Name,omitempty"`
+	Opts       volume.VolumeOpts `json:"Opts,omitempty"`
+	InstanceID string            `json:"Instanceid,omitempty"`
 }
 
 func (mod *Module) Start() error {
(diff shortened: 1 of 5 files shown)


akutz:
Hi @clintonskitson,

Looks like you need to resolve some conflicts. I assigned it to you, but if you want me to handle the resolution please assign it back to me.

clintonskitson:
@akutz rebased, and I believe this is good to go

akutz:
Hi @clintonskitson,

The JSON tags -- JSON field names start with lower-case and Go's encoder can parse between the two.

clintonskitson:
Hmm.. not sure what you're referring to there?  Can you tag the code?  Everything looks to be exported to me.

daemon/module/docker/remotevolumedriver/remvoldriver.go
```diff
@@ -112,8 +112,9 @@ var (
 )
 
 type pluginRequest struct {
```
akutz: Hi @clintonskitson,

In your type definition there are two issues:

1. JSON fields should start with lowercase letters followed by camel case.
2. `ommitempty` should only have one `m`.

The type should be:

```
type pluginRequest struct {
    Name       string            `json:"name,omitempty"`
    Opts       volume.VolumeOpts `json:"opts,omitempty"`
    InstanceID string            `json:"instanceId,omitempty"`
}
```

clintonskitson:
Thanks on #2.  This JSON definition is the receiver as defined by Docker.  #1 is based on Docker's spec here https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md

akutz:
No worries. I wondered if it was in the spec, but thought I'd call it out regardless. Cool cool.

akutz:
Hi @clintonskitson,

One more thing, for any parser to work correctly a branch is either `feature/` or `feature-`. The prefix is always separated by a `/` or a `-`.

PR #25, opened 2015-08-25
REX-Ray Foundation Improvements

akutz:
This patch introduces many improvements to the REX-RAY tool at a core
level in order to lower REX-Ray's barrier of entry to bloggers,
end-users, systems administrators, and software engineers.

All changes introduced by this commit should be considered breaking
changes and updates to existing documentation is required.

INSTALLATION, SYSV, SYSTEMD SUPPORT
REX-Ray now includes built-in support for installing itself as a
service on Linux distributions that support either SystemV or SystemD
initialization systems. This feature has been tested successfully on
both CentOS 7 Minial (SystemD) and Ubuntu 14.04 Server (SystemV)
distributions.

To install REX-Ray on a supported Linux distribution, all that is
required now is to download the binary and execute:

        sudo ./rexray service install

What does that do? In short the above command will determine if the
Linux distribution uses systemctl, update-rc.d, or chkconfig to manage
system services. After that the following steps occur:

  1. The path /opt/rexray is created and chowned to root:root with
     permissions set to 0755.

  2. The binary is copied to /opt/rexray/rexray and chowned to
     root:root with permissions set to 4755. This is important, because
     this means that any non-privileged user can execute the rexray
     binary as root without requiring sudo privileges. For more
     information on this feature, please read about the Linux kernel's
     super-user ID (SUID) bit at
     http://www.tldp.org/HOWTO/Security-HOWTO/file-security.html.

     Because the REX-Ray binary can now be executed with root
     privileges by non-root users, the binary can be used by non-root
     users to easily attach and mount external storage.

  3. The directory /etc/rexray is created and chowned to root:root.

The next steps depends on the type of Linux distribution. However, it's
important to know that the new version of the REX-Ray binary now
supports managing its own PID (at /var/run/rexray.pid) when run as a
service as well as supports the standard SysV control commands such as
"start," "stop," "status," and "restart."

For SysV Linux distributions that use chkconfig or update-rc.d, a
symlink of the REX-Ray binary is created in /etc/init.d/ and then
either "chkconfig rexray on" or "update-rc.d rexray defaults" is
executed.

Modern Linux distributions have moved to SystemD for controlling
services. If the "systemctl" command is detected when installing
REX-Ray then a unit file is written to
/etc/systemd/system/rexray.service with the following contents:

    [Unit]
    Description=rexray
    Before=docker.service

    [Service]
    EnvironmentFile=/etc/rexray/rexray.env
    ExecStart=/opt/rexray/rexray service start
    ExecStart=/opt/rexray/rexray service stop
    KillMode=process
    Restart=always
    StartLimitInterval=0

    [Install]
    WantedBy=docker.service

Additionally, an environment file (and parent directory) is created at
/etc/rexray/rexray.env with the contents:

    REXRAY_HOME=/opt/rexray

The REX-Ray service is not started immediately upon installation. The
install command completes by informing the users that they should visit
http://github.com/emccode/rexray for information on how to configure
REX-Ray's storage drivers. The text to the users also explains how to
start the REX-Ray service once it's configured using the service command
particular to the Linux distribution.

SINGLE SERVICE
This patch also removes the need for REX-Ray to be configured as
multiple service instances in order to provide multiple end-points to
such consumers such as Docker. REX-Ray's backend now supports an
internal, modular design which enables it to host multiple module
instances of any module, such as the DockerVolumeDriverModule. In fact,
one of the default, included modules is...

ADMIN MODULE & REST & USER INTERFACE
The AdminModule enables a RESTful interface for managing REX-Ray's
module system as well as provides a UI to view the currently running
modules. Simply start the REX-Ray server and then visit the URL
http://localhost:7979 in your favorite browser to see what's loaded. Or
you can access either of the currently supported REST URLs:

    http://localhost:7979/r/module/types

and

    http://localhost:7979/r/module/instances

Actually, those aren't the *only* two URLs, but the others are for
internal users as of this point. However, the source *is* open, so... :)

If you want to know what modules are available by using the CLI, after
starting the REX-Ray service simply type:

    [0]akutz@poppy:rexray$ rexray service module types
    [
      {
        "id": 2,
        "name": "DockerVolumeDriverModule",
        "addresses": [
          "unix:///run/docker/plugins/rexray.sock",
          "tcp://:7980"
        ]
      },
      {
        "id": 1,
        "name": "AdminModule",
        "addresses": [
          "tcp://:7979"
        ]
      }
    ]
    [0]akutz@poppy:rexray$

To get a list of the *running* modules you would type:

    [0]akutz@poppy:rexray$ rexray service module instance get
    [
      {
        "id": 1,
        "typeId": 1,
        "name": "AdminModule",
        "address": "tcp://:7979",
        "description": "The REX-Ray admin module",
        "started": true
      },
      {
        "id": 2,
        "typeId": 2,
        "name": "DockerVolumeDriverModule",
        "address": "unix:///run/docker/plugins/rexray.sock",
        "description": "The REX-Ray Docker VolumeDriver module",
        "started": true
      },
      {
        "id": 3,
        "typeId": 2,
        "name": "DockerVolumeDriverModule",
        "address": "tcp://:7980",
        "description": "The REX-Ray Docker VolumeDriver module",
        "started": true
      }
    ]
    [0]akutz@poppy:rexray$

Hmmm, you know, the REX-Ray CLI looks a litle different in the above
examples, doesn't it? About that...

COMMAND LINE INTERFACE
The CLI has also been enhanced to present a more simplified view up
front to users. The commands are now categorized into logical groups:

    [0]akutz@poppy:rexray$ rexray
    REX-Ray:

    Usage:
      rexray [flags]
      rexray [command]

    Available Commands:
      volume      The volume manager
      snapshot    The snapshot manager
      driver      The driver manager
      service     The service controller
      version     Print the version

    Flags:
      -c, --config="/home/akutz/.rexray/config.yaml": The REX-Ray configuration file
      -d, --debug[=false]: Enables verbose output
      -?, --help[=false]: Help for rexray
      -h, --host="tcp://127.0.0.1:7979": The REX-Ray service address

    Use "rexray [command] --help" for more information about a command.

    [0]akutz@poppy:rexray$

For example, this is what happens when the volume command is executed:

    [0]akutz@poppy:rexray$ rexray volume

    Usage:
      rexray volume [flags]
      rexray volume [command]

    Available Commands:
      create      Create a new volume
      get         Get one or more volumes
      map         Print the volume mapping(s)
      path        Print the volume path
      mount       Mount a volume
      attach      Attach a volume
      unmount     Unmount a volume
      detach      Detach a volume
      remove      Remove a volume

    Flags:
      -?, --help[=false]: Help for volume

    Global Flags:
      -c, --config="/home/akutz/.rexray/config.yaml": The REX-Ray configuration file
      -d, --debug[=false]: Enables verbose output
      -h, --host="tcp://127.0.0.1:7979": The REX-Ray service address

    Use "rexray volume [command] --help" for more information about a command.
    [0]akutz@poppy:rexray$

TRAVIS CI SUPPORT
REX-Ray now supports Travis-CI builds either from the primary REX-Ray
repository or via a fork. All builds should be executed through the
Makefile, which is a Travis-CI default. For the Travis-CI settings
please be sure to set the environment variable GO15VENDOREXPERIMENT to
1.


clintonskitson:
Great work!

1. Could `rexray driver` represent `rexray adapter` instead?  I believe this might represent better based on the pattern used.
2. When it initializes multiple adapters (EC2, ScaleIO simultaneously), they are listed as instances.  But the Docker endpoint won't work since there the API calls from Docker only include a volume name.  We might need to tie a service module instance with a specific env variable (DOCKER_STORAGEDRIVERS=blah)?
3. Should we be using `/run/rexray.sock` versus the `127.0.0.1:7979` initially?
4. It looks like it is starting `127.0.0.1:7979` initially, but not sure if this is useful since people don't typically have browsers locally on the distributed systems.  If the web server starts it should be `:7979`?

akutz:
Please note this PR has been updated to include Travis CI support for REX-Ray and that the builds are passing :)

clintonskitson:
Nice!!!!!!!!!!!!!

akutz:
Hi @clintonskitson, "driver" has been renamed "adapter" in the CLI. I will address the other two requests as well.

kacole2:
!!!!!!!! :)

Issue #32, opened 2015-09-10
Add commit number to 'rexray version'


clintonskitson:
Would be nice to directly correlate a commit hash to a binary even with actual releases.  Having `rexray version` return version and commit for all releases would help here.

akutz:
I will be opening a pull request to close this issue.

![image](https://cloud.githubusercontent.com/assets/101085/9803822/97e69cf0-57ec-11e5-9b85-8f8003983052.png)

akutz:
Closed with PR #37

clintonskitson:
Ping @akutz.. In the future I wonder if it is possible to check for uncommitted changes and signify dirty possibly?

akutz:
That's definitely possible, and not that difficult to do. By the way, "Formed" is the build date, not the commit date, so that will help too as it reflects when it was built.

PR #37, opened 2015-09-10
[32] Add commit hash to 'rexray version'

akutz:
This patch handles issue #32 and changes the output of the `rexray version` command to include additional information about the REX-ray version and sources used to build it. For example:

```
[0]akutz@poppy:rexray$ rexray version
SemVer: 0.2.0-rc2+2
Branch: feature/32-version-info
Commit: 86d6ced73b7968cdf292aac11ac6f46c86e01088
Formed: Thu, 10 Sep 2015 18:50:47 CDT
```

Additionally, if the `VERSION` file exists at the root of the repository and the file contains a valid semantic version, then that version's `major`.`minor`.`patch`-`preReleaseTag` components will be used in lieu of those grokked from a call to `git describe --long` at build time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,12 @@ WD := $(shell pwd)
 export MAKEFLAGS := $(MAKEFLAGS) -k
 export GOBIN := $(GOPATH)/bin
 export GO15VENDOREXPERIMENT := 1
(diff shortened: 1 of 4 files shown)


clintonskitson:
very nice

Issue #39, opened 2015-09-11
v0.2 specifying config variables with rexray start doesn't work


kacole2:
@clintonskitson suggested to start rexray by doing `sudo var1=val var2=val rexray service start` to populate the configuration. The configuration is populated, but rexray cannot provision a new volume. 

to reproduce the issue:
```
sudo rexray service install
sudo /etc/init.d/rexray start
sudo rexray service stop
sudo AWS_ACCESS_KEY=mykey AWS_SECRET_KEY=mysecretkey rexray service start
```

```
ubuntu@fun21:~$ rexray volume create --size=20 --volumename="wefw"
FATA[0000] AWS was not able to validate the provided access credentials (AuthFailure)
ubuntu@fun21:~$ rexray service module types
[
  {
    "id": 1,
    "name": "AdminModule",
    "defaultConfigs": [
      {
        "address": "tcp://:7979"
      }
    ]
  },
  {
    "id": 2,
    "name": "DockerRemoteVolumeDriverModule",
    "defaultConfigs": [
      {
        "address": "unix:///run/docker/plugins/rexray-remote.sock",
        "config": {
          "LogLevel": "info",
          "StorageDrivers": [],
          "VolumeDrivers": [
            "docker"
          ],
          "OsDrivers": [
            "linux"
          ],
          "MinVolSize": 0,
          "RemoteManagement": false,
          "DockerVolumeType": "",
          "DockerIops": 0,
          "DockerSize": 0,
          "DockerAvailabilityZone": "",
          "AwsAccessKey": "mykey",
          "AwsSecretKey": "mysecretkey",
          "AwsRegion": "",
          "RackspaceAuthUrl": "",
          "RackspaceUserId": "",
          "RackspaceUserName": "",
          "RackspacePassword": "",
          "RackspaceTenantId": "",
          "RackspaceTenantName": "",
          "RackspaceDomainId": "",
          "RackspaceDomainName": "",
          "ScaleIoEndpoint": "",
          "ScaleIoInsecure": false,
          "ScaleIoUseCerts": true,
          "ScaleIoUserName": "",
          "ScaleIoPassword": "",
          "ScaleIoSystemId": "",
          "ScaleIoSystemName": "",
          "ScaleIoProtectionDomainId": "",
          "ScaleIoProtectionDomainName": "",
          "ScaleIoStoragePoolId": "",
          "ScaleIoStoragePoolName": "",
          "XtremIoEndpoint": "",
          "XtremIoUserName": "",
[...]

Issue #38, opened 2015-09-11
v0.2 put ENV variables into service module types


kacole2:
rexray still functions in v0.2 just as it used to in v0.1x when ENV variables were used for configuration. I can still declare ENV variables and rexray still creates volumes. However, the ENV variables are not populated in the configuration output with `rexray service module types`.

akutz:
Hi @kacole2,

It appears this is a duplicate of #39 and so I'm closing this ticket.

PR #41, opened 2015-09-11
[40] Always load /etc/environment if present

akutz:
Because sudo doesn't always play nice with /etc/environment and so many people place configuration information there even when inappropriate or not fully understood, REX-Ray should manually load /etc/environment in order to relieve users of possible issues due to Linux systems being so disparate and supporting so many different configuration options.

This patch solves the above issue. Although tangential, issues #38 and #39 are related.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


clintonskitson:
Makes a ton of sense

Issue #43, opened 2015-09-11
Add an example config.json/config.yaml file


clintonskitson:
In order to simplify how someone would figure out how to leverage RR, we should include an example file or include it in the README.  @kacole2

kacole2:
you're right... someone should :+1:

PR #44, opened 2015-09-11
added yaml example config file

clintonskitson:
This patch adds an example config yaml file.

### DIFF
--- a/examples/config.yaml
+++ b/examples/config.yaml
@@ -1,0 +1,4 @@
+StorageDrivers: 
+  - ec2
+AwsAccessKey: 
+AwsSecretKey: 


clintonskitson:
Referencing issue https://github.com/emccode/rexray/issues/43

akutz:
Hi @clintonskitson,

I'd like to close this PR without accepting it so I can create a default config file that includes all of the default options as well as commented out provider options with explanations. Thoughts?

kacole2:
one thing to add is to please note where the `config.yml` file needs to be dropped so RR can read it. or note if it must be referenced during `sudo rexray service start`

thanks gents

akutz:
Hi @kacole2,

The command already shows one place:

```
REX-Ray:

Usage: 
  rexray [flags]
  rexray [command]

Available Commands: 
  volume      The volume manager
  snapshot    The snapshot manager
  adapter     The adapter manager
  service     The service controller
  version     Print the version

Global Flags:
  -c, --config="/home/akutz/.rexray/config.yaml": The REX-Ray configuration file
  -?, --help[=false]: Help for rexray
  -h, --host="tcp://:7979": The REX-Ray service address
  -l, --logLevel="info": The log level (panic, fatal, error, warn, info, debug)
  -v, --verbose[=false]: Print verbose help information

Use "rexray [command] --help" for more information about a command.
```

However, I'll update the docs to reflect the official locations. FWIW, anyone can open a PR and update the docs. :)

kacole2:
thanks... i didn't notice that. 

as for the hidden directory `/home/akutz/.rexray/`, was that created during `sudo rexray service install`? Just curious because I will need to retrieve the home path for the current user and append that folder for the docker-machine process.

akutz:
Hi @kacole2,

It doesn't exist. The rule is that the config file is parsed first from $HOME/.rexray/config.yaml if it exists and then /etc/rexray/config.yaml if that exists. Otherwise it doesn't use a config file unless one is specified on the command line with the flag you see above.

kacole2:
Perfect.  Thanks

clintonskitson:
I'm good with that @akutz

PR #45, opened 2015-09-11
updated README to include new config options

clintonskitson:
This patch brings the CLI information towards the top of the readme and adds some information about env and config based approaches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,16 @@ This might currently require upstream additions for the Goamz package to github.
 docker run --rm -it -e GO15VENDOREXPERIMENT=1 -v $GOPATH:/go -w /go/src/github.com/emccode/rexray/ golang:1.5 make install
 ```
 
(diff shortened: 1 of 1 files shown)


akutz:
Hi @clintonskitson,

There's a typo in the updated README.

```
There are two methods to configure REX-Ray.  One is through usage of 
environment variables and the other is through configuration files.  The other 
is through configuration files.  These two methods are respected equally 
whether interacting through the CLI or packages/library.
```

Note that the third sentence repeats the second part of the second sentence. Also, you omitted the CLI analogues. Should we have an additional column where we list the values from the constants file and indicate these are the keys for the config files or CLI flags?

clintonskitson:
Fixed typo.  Removed part that referenced examples, since that other PR isn't in yet.  Not sure what you mean around `CLI analogues`?

PR #69, opened 2015-09-20
Unstable Repo, Installer, Package Support, REXRAY_HOME, and Service Logging

akutz:
This patch combines five changes:

1. The `stupid` repository has been renamed to `unstable`
2. A standard installation/uninstallation system has been created that is used regardless whether executed manually or by a package manager.
3. Support for building and deploying RPM and DEB packages
4. The `REX-Ray` binary may live wherever, but all of the data is created at the location defined by the environment variable `REXRAY_HOME`.
5. When executed as a service REX-Ray logs to `$REXRAY_HOME/var/log/rexray.log`

## Repositories
There's not much more to say but that the repository once known as `stupid` has been renamed to `unstable`. The old `stupid` links will continue to work for versions pushed to the repository when it was named such, but once the next release candidate is pushed, those versions will be purged.

## Installer
The `REX-Ray` binary now supports the `install` and `uninstall` commands, executed as `rexray install` or `rexray uninstall`. These commands install the `REX-Ray` service if on a Linux system. Additionally, the `install` and `uninstall` commands are also what is used by the RPM and DEB packages, meaning that the `REX-Ray` installer is distribution and package agnostic.

## Package Support
The Makefile now has support for the following targets:

* rpm
* rpm-all
* rpm-linux-amd64
* rpm-linux-386
* deb
* deb-all
* deb-linux-amd64

The above targets do what they imply they do -- they assemble the REX-Ray binary into an RPM or DEB package. Please note the above targets expect the appropriate architecture of REX-Ray to already be built and will not force a new build of REX-Ray.

Also, it's not without *much* trouble that building a 32-bit DEB package is possible on a 64-bit system, and thus the 32-bit DEB package is not available at this time. Since `alien` was used to create the DEB packages from the RPMS, the 32-bit RPM could easily be converted to a 32-bit DEB if a 32-bit Debian system were available.

The packages install `REX-Ray` and register its service, but the server is not started.

## `REXRAY_HOME`
The `REX-Ray` binary can exist anywhere on a system, but when it is executed it assumes the following paths (and will create them if they do not exist):

* `$REXRAY_HOME/etc/rexray`
* `$REXRAY_HOME/var/lib/rexray`
* `$REXRAY_HOME/var/log/rexray`
* `$REXRAY_HOME/var/run/rexray`

If the `REXRAY_HOME` environment variable does not exist then the above paths will exist within the filesystem layout standard for both almost any UNIX-derived system such as Linux or Darwin (OS X).

[...]

PR #70, opened 2015-09-21
Patch xtremio storage driver package name

clintonskitson:


volume/volume.go
```diff
@@ -87,6 +88,11 @@ func getDrivers(
 
 func (vdm *VolumeDriverManager) Mount(volumeName, volumeID string, overwriteFs bool, newFsType string) (string, error) {
 	for _, driver := range vdm.Drivers {
+		log.WithFields(log.Fields{
+			"volumeName":  volumeName,
+			"volumeID":    volumeID,
+			"overwriteFs": overwriteFs,
+			"newFsType":   newFsType}).Infof("%s: Mount()", driver.Name())
```
akutz: Hi @clintonskitson,

Please update the logging calls in this PR in the same manner per the comment below:

> I'm only going to comment on this once for this file, but it's applicable in numerous places. The `Debugf` and all the `LOGf` functions in logrus exist for drop-in compatibility with `stdlib`'s `log` package only. All places where the format variants are used should be changed. For example, the above line should look like this:

>```
>        log.WithFields(log.Fields{
>            "volumeName":  volumeName,
>            "volumeID":    volumeID,
>            "overwriteFs": overwriteFs,
>            "newFsType":   newFsType,
>            "driverName": driver.Name()}).Debu("mounting volume")
>```
>
> However, if you'll look inside the providers, I took great pains to log most everything at the provider level as possible as all log statements always include the providerName field in conjunction with that provider's constant value denoting the same.

clintonskitson:
@akutz My fault, that was a bad PR.  This should have the relevant changes.

akutz:
Hi @clintonskitson,

I'm holding off on merging this until PR #69 is approved so the build as a result of this merge will be uploaded to the `unstable` repo as well as included in an RPM and DEB. Other than that, no reason to not approve it.

PR #94, opened 2015-09-28
added create from existing volume or snap as option

clintonskitson:
This PR enables the ability to specify volumename, volumeid, snapshotname, snapshotid as volume options through the Docker vdm.  If these options are specified, the new volume will be created from an existing volume.


clintonskitson:
Could be useful, but let's address more detailed logging standardly for the package in another PR.

PR #96, opened 2015-09-29
Updated logging on docker daemon module

clintonskitson:
This ensures that errors are logged to the logging package
when the Docker volumedriver module is called.  This is
in addition to the standard HTTP logging.

### DIFF
--- a/daemon/module/docker/volumedriver/voldriver.go
+++ b/daemon/module/docker/volumedriver/voldriver.go
@@ -2,7 +2,6 @@ package volumedriver
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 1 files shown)


daemon/module/docker/volumedriver/voldriver.go
```diff
@@ -212,12 +213,14 @@ func (mod *Module) buildMux() *http.ServeMux {
 		var pr pluginRequest
 		if err := json.NewDecoder(r.Body).Decode(&pr); err != nil {
 			http.Error(w, fmt.Sprintf("{\"Error\":\"%s\"}", err.Error()), 500)
+			log.Error(errors.WithError("/VolumeDriver.Create: error decoding json", err))
```
akutz: Hi @clintonskitson,

These should be `log.WithField("error", err).Error("/VolumeDriver.Create: error decoding json")`. I won't repeat this for the entire file, but it's in numerous places.

clintonskitson: Duh late night, tganks

clintonskitson:
@akutz thanks for the eyes, this one looks to be ready now

Issue #50, opened 2015-09-14
Hide specific configuration parameters from logging and return


clintonskitson:
Since the new foundational enhancements with an admin API and logging, we now need think about hiding configuration parameters in JSON conversion and logging.  

Considering the default behavior is to expose TCP admin session, a simple call like `http://ip:7979/r/module/instances` is going to expose too much information.  I would assume that returning this kind of information upon instantiation may be appropriate, but that returning it after that point should not be possible.

```
[
  {
    "id": 1,
    "typeId": 1,
    "name": "AdminModule",
    "config": {
      "address": "tcp://:7979"
    },
    "description": "The REX-Ray admin module",
    "started": true
  },
  {
    "id": 2,
    "typeId": 2,
    "name": "DockerRemoteVolumeDriverModule",
    "config": {
      "address": "unix:///run/docker/plugins/rexray-remote.sock",
      "config": {
        "LogLevel": "info",
        "StorageDrivers": [
          "ec2"
        ],
        "VolumeDrivers": [
          "docker"
        ],
        "OsDrivers": [
          "linux"
        ],
        "MinVolSize": 0,
        "RemoteManagement": false,
        "DockerVolumeType": "",
        "DockerIops": 0,
        "DockerSize": 0,
        "DockerAvailabilityZone": "",
        "AwsAccessKey": "A",
        "AwsSecretKey": "jSv",
```

akutz:
Hi @clintonskitson,

Do you have the link you sent me a few weeks back on Slack about HTTP over the UNIX socket implementation?

clintonskitson:
David does a good job of receiving the tcp/unix sockets here.
https://github.com/calavera/dkvolume/blob/master/api.go

And this is the Docker client making the calls to that API.
https://github.com/docker/docker/blob/master/pkg/plugins/client.go

clintonskitson:
If you're thinking that a fix is to leverage the unix socket and file permissions then I would agree with the approach.  Realistically those creds are stored in files and in the environment, so I don't see any extra security problems if the TCP isn't being exposed.

akutz:
Hi @clintonskitson,

I'm thinking the fix is both approaches. That's in part why I invented Golf. Also, so I could say I created Golf :)

clintonskitson:
Next think you're gonna make the *Internet* =)

PR #100, opened 2015-09-29
Config Data is Now Secure & Plain-Text

akutz:
This patch is in response to issue #50. The configuration data in REX-Ray is now stored separately as secure & plain-text. The plain-text version of the data is serialized and sent to the server when necessary to provide credentials. However, when executing commands like `rexray service module types`, the produced output will be secure by default and no longer include the following configuration keys:

  * AwsSecretKey
  * RackspacePassword
  * ScaleIoPassword
  * XtremIoPassword

For example:

    [0]akutz@poppy:rexray$ rexray service module types
    [
      {
        "id": 1,
        "name": "AdminModule",
        "defaultConfigs": [
          {
            "address": "tcp://:7979"
          }
        ]
      },
      {
        "id": 2,
        "name": "DockerRemoteVolumeDriverModule",
        "defaultConfigs": [
          {
            "address": "unix:///run/docker/plugins/rexray-remote.sock",
            "config": {
              "LogLevel": "info",
              "StorageDrivers": [],
              "VolumeDrivers": [
                "docker"
              ],
              "OsDrivers": [
                "linux"
              ],
              "MinVolSize": 0,
              "RemoteManagement": false,
              "DockerVolumeType": "",
              "DockerIops": 0,
              "DockerSize": 0,
              "DockerAvailabilityZone": "",
              "AwsAccessKey": "",
              "AwsRegion": "",
              "RackspaceAuthUrl": "",
              "RackspaceUserId": "",
              "RackspaceUserName": "",
              "RackspaceTenantId": "",
              "RackspaceTenantName": "",
              "RackspaceDomainId": "",
              "RackspaceDomainName": "",
              "ScaleIoEndpoint": "",
              "ScaleIoInsecure": false,
              "ScaleIoUseCerts": true,
              "ScaleIoUserName": "",
              "ScaleIoSystemId": "",
              "ScaleIoSystemName": "",
              "ScaleIoProtectionDomainId": "",
              "ScaleIoProtectionDomainName": "",
              "ScaleIoStoragePoolId": "",
              "ScaleIoStoragePoolName": "",
              "XtremIoEndpoint": "",
              "XtremIoUserName": "",
              "XtremIoInsecure": false,
              "XtremIoDeviceMapper": "",
              "XtremIoMultipath": "",
              "XtremIoRemoteManagement": false
            }
          }
        ]
      },
      {
        "id": 3,
        "name": "DockerVolumeDriverModule",
        "defaultConfigs": [
          {
            "address": "unix:///run/docker/plugins/rexray.sock",
            "config": {
              "LogLevel": "info",
              "StorageDrivers": [],
              "VolumeDrivers": [
                "docker"
              ],
              "OsDrivers": [
                "linux"
              ],
              "MinVolSize": 0,
              "RemoteManagement": false,
              "DockerVolumeType": "",
              "DockerIops": 0,
              "DockerSize": 0,
              "DockerAvailabilityZone": "",
              "AwsAccessKey": "",
              "AwsRegion": "",
              "RackspaceAuthUrl": "",
              "RackspaceUserId": "",
              "RackspaceUserName": "",
              "RackspaceTenantId": "",
              "RackspaceTenantName": "",
              "RackspaceDomainId": "",
              "RackspaceDomainName": "",
              "ScaleIoEndpoint": "",
              "ScaleIoInsecure": false,
              "ScaleIoUseCerts": true,
              "ScaleIoUserName": "",
              "ScaleIoSystemId": "",
              "ScaleIoSystemName": "",
              "ScaleIoProtectionDomainId": "",
              "ScaleIoProtectionDomainName": "",
              "ScaleIoStoragePoolId": "",
              "ScaleIoStoragePoolName": "",
              "XtremIoEndpoint": "",
              "XtremIoUserName": "",
              "XtremIoInsecure": false,
              "XtremIoDeviceMapper": "",
              "XtremIoMultipath": "",
              "XtremIoRemoteManagement": false
            }
          }
        ]
      }
    ]

The data is still available on the server-side in the same locations, it's just not serialized for transport unless specifically requested.

I would ask @clintonskitson, @kacole2, and @jonasrosland to please review the fields that I categorized as *secure* and note whether any others should be placed into that category as well.


kacole2:
that's really awesome! Nice thinking :+1:

clintonskitson:
@akutz `AwsAccessKey` should be a secure key as well

kacole2:
@clintonskitson the AccessKey is like the ScaleIO username. does it really need to be secure?

akutz:
Hi @kacole2,

That was my thinking as well, but that's one of the fields that caused me to reach out for feedback as I just don't have as much experience with it as you or @clintonskitson.

clintonskitson:
Good question, my dunce hat tells me that it is important because it's not in the username form.  But I was just assuming based on previous experiences.  Need to look further..

clintonskitson:
@drumulonimbus  ?

akutz:
Amazon doesn't think it's super secret, just to not be an idiot with it... http://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html

clintonskitson:
I am actually inclined to make the username and password secret for all plugins.

akutz:
I guess I view it as an SSH pub and private key, or a cert pair. The public side is designed to be...well...public. And it may be useful for the server to return the Access Key (user ID) to the user.

akutz:
Hi @clintonskitson,

I disagree for identification purposes as I just outlined.

drumulonimbus:
I'm on the fence here - one side of me thinks that *all* account identifiers should be secret, but I wonder if that might make things actively more difficult for the end user.

clintonskitson:
Ok, let's leave it as is.  If someone can actually describe a good reason why we would then we can re-address easily.

ryandotclair:
I'd actually reverse that @clintonskitson. What's the cost/downside of doing it? If there really isn't any, and it's "cheap" to implement, I say do it. More paranoid the better, so long as it doesn't hamper anything.

akutz:
Hi @ryandotclair,

The downside is diagnostics. When I ask the system to tell me which modules are loaded, I will have zero idea of any configured credentials for those modules. I'm not speaking of the passphrases or private keys, but simply the user names. We must consider how to diagnose issues when they occur, and a very early question is going to be, what credentials are being used to connect to X?

ryandotclair:
To me it sounds like we've got the answer then. Don't do it :smile:

akutz:
Hi @ryandotclair,

I disagree, it is done. The user names at least. Secret credentials are hidden. We can revisit this issue in a future release.

Issue #111, opened 2015-10-09
Retrieve instance UUID in Openstack


maricaantonacci:
Using rackspace driver results in an error while getting instance uuid with /usr/bin/xenstore-read
The UUID may be retrieved using command /usr/sbin/dmidecode:
```
Handle 0x0100, DMI type 1, 27 bytes
System Information
	Manufacturer: OpenStack Foundation
	Product Name: OpenStack Nova
	Version: 2014.2.3
	Serial Number: 47513848-0036-2500-905b-0025905bca52
	UUID: 70671C29-1E7A-48BC-BBBB-24AC5CFF0DE9
	Wake-up Type: Power Switch
	SKU Number: Not Specified
	Family: Not Specified
```

clintonskitson:
I have created a special branch to help troubleshoot this.

Download `rexray` from here.  Make sure to run it with `./rexray` for now.  
https://drive.google.com/file/d/0B6vlRdQM169ucm4xWDloVDVhVGM/view?usp=sharing

I introduced new parameters in this patchfor Openstack versus the previous Rackspace ones.  Please also indicate `openstack` as a storage driver.
```
      --openstackAuthUrl="": 
      --openstackDomainId="": 
      --openstackDomainName="": 
      --openstackPassword="": 
      --openstackTenantId="": 
      --openstackTenantName="": 
      --openstackUserId="": 
      --openstackUserName="": 
```

maricaantonacci:
It fails. I think because of upper case. The command should be:

/usr/sbin/dmidecode | awk '/UUID:/ {print tolower($2)}' 

The error message:
DEBU[0000] error initializing driver                     driverName=Openstack error.provider=Openstack error.cmd.Args=[/usr/sbin/dmidecode | awk '/UUID:/ {print $2;}'] error.cmd.Out=[] error.cmd.Path="/usr/sbin/dmidecode | awk '/UUID:/ {print $2;}'" error.message="error getting instance id"

clintonskitson:
Ok, I see the problem, will send a new binary.

clintonskitson:
Try this one

https://drive.google.com/file/d/0B6vlRdQM169uNmVGb3l2VWFYWjQ/view?usp=sharing

maricaantonacci:
I see again /usr/bin/xenstore-read...

<pre>
DEBU[0000] updated log level                             logLevel=debug
[...]

PR #116, opened 2015-10-12
Update rackspace package for config params

clintonskitson:
The foundational changes updated configuration per package.  The
Rackspace package had a regression introduced where the contsants
from the config package were being used instead of the configured
cfg values.


clintonskitson:
This satisfies the `rackspace` driver fix from https://github.com/emccode/rexray/issues/111

Issue #129, opened 2015-10-16
Creating a volume from an existing volume by default sets 16GB size


clintonskitson:
When creating a new volume from an existing volume, currently the results depend on the provider.  For EC2, it allows arbitrary settings for size, volume type, etc.  Currently, if you don't specify these settings, default values will be assumed.

Consistent behavior for this between driver would be ideal, where initially at least replicating the settings within the new volume would take place.  Otherwise, it should be left to the driver to determine if they respect custom settings that are provided.

akutz:
Hi @clintonskitson,

Would you like help on this? If so, can you describe the issue in more detail? Thanks!

Issue #124, opened 2015-10-15
amazon: support volume mount across availability zones


jaroslavl1:
In amazon volumes cannot be mounted on ec2 instances in different availability zones (I assume same region). This leads to error message like this:

core@ip-10-60-34-175 ~ $ sudo -E rexray volume mount --volumename="vol4"
INFO[0000] mounting volume                               driverName=docker newFsType= overwriteFs=false volumeID= volumeName=vol4
INFO[0000] mounting volume                               driverName=docker newFsType= overwriteFs=false volumeID= volumeName=vol4
INFO[0000] Got next device name: /dev/xvdc
FATA[0000] The volume 'vol-acb2714a' is not in the same availability zone as instance 'i-02b49cc4' (InvalidVolume.ZoneMismatch)

however it can be done by creating volume snapshot and then creating volume from it and deleting the original volume - thus moving it to different availability zone. This can take some time in case of bigger volumes. It should be automated by rexray.

This issue makes it impossible to use Docker rexray volume driver on machines in different AZs since it cannot mount volumes from other AZ.

clintonskitson:
@jaroslavl1 I can think of two ways REX-Ray can accomplish this for you.  Below are the flags for a `create` operation.

```
Flags:
      --availabilityzone="": availabilityzone
      --iops=0: IOPS
      --runasync[=false]: runasync
      --size=0: size
      --snapshotid="": snapshotid
      --volumeid="": volumeid
      --volumename="": volumename
      --volumetype="": volumetype
```

Notice how there is an `availabilityzone` flag.  If you combine this with either a `volumeid` or a `snapshotid` it will perform the workflow you specified to create a new volume in another or the local AZ to the instance.  If you are in a new AZ and encountering an error that you specify, then you don't need to specify the `availabilityZone` option.  But if you create a volume from an instance in the same AZ as the existing volume then you need to specify the `availabilityZone` option.

jaroslavl1:
I was successful by following your suggestion:

```
rexray volume create --volumeid="vol-9b7fc85a" --volumename="vol2" --volumetype="gp2"
```

The --availabilityzone="us-west-2b" parameter wasn't even necessary in my case and rexray recognized the correct availability zone. The result was as expected - new volume created with the name as specified by first creating snapshot, then volume from snapshot then deleting snapshot.

One suggestion I have here is by default to use the same volumetype and iops as the original volume. Without volumetype parameter I got the standard magnetic volume.

[...]

PR #142, opened 2015-10-19
Set default settings for new volumes from existing are from orig

clintonskitson:
This commit ensures that volumes created from existing snapshots
or volumes are configured with the same defaults.  For volumes,
the VolumeType, IOPS, and Size are replicated.  For snaphshots,
the Size is replicated.  The parameters to override this are
standard volume creation options.


clintonskitson:
Satisfies feedback from https://github.com/emccode/rexray/issues/124 and https://github.com/emccode/rexray/issues/129

PR #144, opened 2015-10-20
Bugfix/scaleio get snapshots

clintonskitson:
This PR fixes a problem where ScaleIO snapshots were not being returned.


clintonskitson:
fyi @akutz

PR #165, opened 2015-10-26
Updated config fields from rackspace

clintonskitson:
The Openstack driver has been updated to change config fields
from Rackspace to Openstack.  The change will be important
moving forward but didn't impact the existing driver.


akutz:
Hi @clintonskitson,

I merged the PR. In the future, if you could please remember to add the label, milestone, etc., I'd appreciate it :)

PR #174, opened 2015-10-29
Config Refactor - Dynamic Config Keys

akutz:
This patch updates the Config class so that all configuration keys, command-line flags, environment variables, etc. are defined by their consumers. The effect of this is that there are no more static keys as class-level fields on the Config class. Consumers of config values must access them using the Config class's Viper object.


akutz:
Hi @clintonskitson,

Please review this when you have a chance. It's got beaucoup tests and ready for commit.

PR #187, opened 2015-11-30
Fix for build issue due to Viper API change

akutz:
This patch fixes the build error due to the Viper API changing from ReadConfigNoNil to MergeConfig.


akutz:
I'm closing this PR because I realize now it's invalid due to the testing changes. I'll open another PR integrating Gofig instead.

Issue #186, opened 2015-11-30
Compilation failure on OSX


gijzelaerr:
When I try to compile current rexray on OSX I fail:
```
λ  cd rexray
λ  make
SemVer: 0.3.0-dev+10
RpmVer: 0.3.0+dev+10
Branch: master
Commit: b0729581afb9bbc8647e3a1fab558a16efc4ab01
Formed: Mon, 30 Nov 2015 11:22:49 SAST

target: deps
  ...installing glide...SUCCESS!
  ...downloading go dependencies...SUCCESS!
target: fmt
  ...formatting rexray...SUCCESS!
target: install
  ...installing rexray Darwin-x86_64...# github.com/emccode/rexray/core/config
core/config/config.go:202: c.v.ReadConfigNoNil undefined (type *viper.Viper has no field or method ReadConfigNoNil)
make: *** [_install] Error 2
make: Target `all' not remade because of errors.
λ  git describe
v0.2.1-10-gb072958
```

Unfortunately i'm absolutely unfamiliar with go, but please let me know if you need more debug information.

akutz:
Hi @gijzelaerr,

This is a bug due to a dependency change. I'll be pushing a PR to accommodate this shortly.

akutz:
Hi @gijzelaerr,

PR #188 has been filed that fixes this issue.

gijzelaerr:
i can confirm this has solved the problem, thanks!

akutz:
Hi @gijzelaerr,

You're very welcome. Please let me know if you run into any other issues :)

PR #188, opened 2015-11-30
Gofig Integration / Namespace Change

akutz:
This patch does away with the core/config package and instead uses the Gofig project for all of the configuration responsibilities. Additionally all of the REX-Ray properties are now under the rexray namespace in the YAML configuration. What was once "host" is now "rexray.host." These changes are reflected in the documentation.

This also handles issue #186.


akutz:
Hi @clintonskitson,

Please test this PR's build against some real data and let me know whether it's safe to merge. Thanks!

clintonskitson:
@akutz 
```
DEBU[0000] invoking service start                        os.Args=[./rexray start -l debug]
Starting REX-Ray...DEBU[0000] trying to start service                       client=/tmp/uwrVGTJxstSECXhggEqghJBXDpmSQXnd.sock
Start the service

Usage: 
  rexray service start [flags]

Flags:
      --client="": Socket the daemon uses to communicate to the client
      --force[=false]: Forces the service to start, ignoring errors
  -f, --foreground[=false]: Starts the service in the foreground

Global Flags:
  -c, --config="": The path to a custom REX-Ray configuration file
  -?, --help[=false]: Help for start
  -h, --rexrayHost="tcp://:7979": The REX-Ray host
  -l, --rexrayLogLevel="warn": The log level (error, warn, info, debug)
  -v, --verbose[=false]: Print verbose help information


Error: unknown flag: --logLevel
```

PR #190, opened 2015-12-02
Added pre-emption capabilities to drivers

clintonskitson:
There is now the ability to pre-empt on mounts through the CLI
as well as directly through HTTP/JSON API.  In the case of CLI,
this in invoked through force with attach/detach and preempt with
mount.  Additionally, pre-emption can be enabled globally through
the volume.mountpreempt configuration flag.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -256,3 +256,33 @@ supported:
 Docker   | docker
 
 The volume driver `docker` is automatically activated.
+
+### Pre-emptive Volume Mount
+There is a capability to pre-emptively detach any existing attachments to other
+instances before attempting a mount.  This will enable use cases for
+availability where another instance must be able to take control of a volume
+without the current owner instance being involved.  The operation is considered
+equivalent to a power off of the existing instance for the device.
+
+Example configuration file follows:
+```yaml
+rexray:
+  storageDrivers:
+  - openstack
+  volume:
+    mountPreempt: true
+openStack:
+  authUrl: https://authUrl:35357/v2.0/
+  username: username
+  password: password
+  tenantName: tenantName
+  regionName: regionName
+```
+
+Driver|Supported
+------|---------
+EC2|Yes
+OpenStack|With Cinder v2
+ScaleIO|Yes
+Rackspace|No
+XtremIO|Yes
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -270,7 +270,8 @@ rexray:
   storageDrivers:
   - openstack
   volume:
-    mountPreempt: true
+    mount:
+      preempt: true
 openStack:
   authUrl: https://authUrl:35357/v2.0/
   username: username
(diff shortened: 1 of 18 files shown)


.docs/user-guide/config.md
```diff
@@ -256,3 +256,22 @@ supported:
...
+### VOLUME MOUNT PRE-EMPTION
+There is a capability to pre-emptively detach any existing attachments to other
+instances before attempting a mount.  This will enable use cases for
+availability where another instance must be able to take control of a volume
+without the current owner instance being involved.  The operation is considered
+equivalent to a power off of the existing instance for the device.
+
+Property Name | Environment Variable | CLI Flag
+--------------|----------------------|-------------
+`volume.mountpreempt`   | `REXRAY_VOLUME_MOUNTPREEMPT`          | `--preempt`
```
akutz: Hi @clintonskitson,

Wouldn't the property name and CLI flag be `rexray.volume.mountpreempt` and `rexrayPreempt` (respectively)? Also, if you note, I actually no longer use the tables elsewhere to show variations of the name but reference the Configuration section with a link so readers can see how the names are derived from a single property name.

.docs/user-guide/config.md
```diff
@@ -256,3 +256,22 @@ supported:
...
+
+Driver|Supported
+------|---------
+EC2|Yes
+OpenStack|With Cinder v2
+ScaleIO|Yes
+Rackspace|No
+XtremIO|Soon
```
akutz: https://youtu.be/356faqb9JnU

.docs/user-guide/config.md
clintonskitson: So I was going back and forth about the naming of the param, and hence ended up in something that doesn't make sense =)

Two things 

1) I see the param being used from the CLI in one way now, with the mount command.  Does it make sense to add on `rexray` to that? ie. `rexray mount --volumename=test --preempt`.  I could see it as `--volumepreempt` as well though.

2) I went back and forth about where to place it in the config file.  At one point I was considering it a driver level configurable item.  But it is now implemented at the vdm level, so probably makes sense the way you list it as a RR context setting for volumes.

akutz: Hi @clintonskitson,

If it's a gofig-defined flag, which it's...it's not. So if it doesn't use gofig then you shouldn't list the property name OR environment variable name as they're not defined. It appears you're *only* defining it as a flag.

clintonskitson: Not sure I am following what you're saying completely.

How would you rather see it if it's a vdm defined flag?

```
---
rexray:
  storageDrivers:
  - openstack
openStack:
  authUrl: url
  username: clintonskitson
  password: pass
  tenantName: 10677883784544-Project
  regionName: region-a.geo-1
volume:
  mountPreEmpt: true
```

```
---
rexray:
  storageDrivers:
  - openstack
  volume:
    mountPreEmpt: true
openStack:
  authUrl: url
  username: clintonskitson
  password: pass
  tenantName: 10677883784544-Project
  regionName: region-a.geo-1
```

clintonskitson:
@akutz I just updated this PR based on our conversation

akutz:
Hi @clintonskitson,

I'm reviewing it now.

PR #191, opened 2015-12-02
Fix for not setting Gofig global/user dirs

akutz:
This patch fixes the issue where the Gofig's global (/etc) and user ($HOME) directories were not set, resulting in the default config files not being loaded.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -1,11 +1,19 @@
 package core
 
 import (
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


akutz:
Hi @clintonskitson,

Please test this to see if it fixes the issue you identified last night regarding the configuration files not being loaded properly. Thanks!

PR #209, opened 2015-12-12
Added VirtualBox driver

clintonskitson:
This commit adds a VirtualBox driver.  The driver enables standard
functionality when running RR inside of VirtualBox drivers.
Snapshots are not supported.

### DIFF
--- a/drivers/storage/virtualbox/virtualbox.go
+++ b/drivers/storage/virtualbox/virtualbox.go
@@ -1,0 +1,678 @@
+package virtualbox
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/akutz/gofig"
+	"github.com/akutz/goof"
+	vbox "github.com/appropriate/go-virtualboxclient/virtualboxclient"
+
+	"github.com/emccode/rexray/core"
+	"github.com/emccode/rexray/core/errors"
+)
+
+const providerName = "virtualbox"
+
+type driver struct {
+	virtualbox *vbox.VirtualBox
+	machine    *vbox.Machine
+	r          *core.RexRay
+	m          sync.Mutex
+}
+
+func ef() goof.Fields {
+	return goof.Fields{
+		"provider": providerName,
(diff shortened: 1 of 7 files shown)


drivers/storage/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,625 @@
...
+func (d *driver) controllerName() string {
+	cn := d.r.Config.GetString("virtualbox.controllerName")
+	if cn == "" {
+		return "SATA Controller"
+	}
+	return cn
+}
+
+func configRegistration() *gofig.Registration {
+	r := gofig.NewRegistration("XtremIO")
```
akutz: Hi @clintonskitson,

Please correct this to read "VirtualBox".

drivers/storage/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,625 @@
...
+	log "github.com/Sirupsen/logrus"
+	"github.com/akutz/gofig"
+	"github.com/akutz/goof"
+	"github.com/appropriate/go-virtualboxclient/virtualboxclient"
+
+	"github.com/emccode/rexray/core"
+	"github.com/emccode/rexray/core/errors"
+)
+
+const providerName = "VirtualBox"
```
akutz: Hi @clintonskitson,

All the provider names get lcased anyway, so I'm not sure if we want to continue to describe them in the constants with mixed-case as later comparisons may be in err. Also, may I suggest `vbox` for brevity? I think they use that internally.

drivers/storage/virtualbox/virtualbox.go
```diff
@@ -0,0 +1,625 @@
...
+	"net"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/akutz/gofig"
+	"github.com/akutz/goof"
+	"github.com/appropriate/go-virtualboxclient/virtualboxclient"
```
akutz: Hi @clintonskitson,

For brevity you may wish to import this as `vbox` or `vboxc`.

akutz:
Hi @clintonskitson,

Other than the in-line code comments, this looks good. Please review/update where appropriate and then we can merge!

Issue #171, opened 2015-10-29
Create Volume locking mechanism


clintonskitson:
Currently REX-Ray expects that the software calling it maintains and is the authority about the locking mechanism for Volumes.  If REX-Ray is asked to unmount, then it accepts the unmount command.  If multiple containers have a volume mounted, REX-Ray is unaware of this since it has no context of containers, and only amount of requests per mount.

A solution was to include this in Docker with the following PR, but was rejected.

https://github.com/docker/docker/pull/15511

To do this properly where the responsibility is jointly handled, the Docker Volume Driver API needs to be enhanced.  The reason for this is that an exchange of information must take place between the Docker Daemon and the Volume Driver about the running containers.  If this exchange does not happen, then if either service resets then the accounting on either side will be invalid and cause issues.

I don't believe the right solution is a locking mechanism in RR, but for now we should include something that is configurable and non-persistent (relying on accuracy through staying active).

1) count the mounts for a volume
2) actually perform unmount when unmount received and mount count == 1

PR #212, opened 2015-12-13
Added volume mount/unmount accounting and control for unmounts

clintonskitson:
This commit introduces an accounting mechanism at the volume
driver manager layer.  It is will always perform counting
of create/mount/unmount operations that are received through
the HTTP/JSON API.  If the rexray.volume.unmount.checkusedcount
is set to true, then it will stop unmounts since it is detecting
that another container is using the mount still.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -259,7 +259,8 @@ Docker   | docker
 
 The volume driver `docker` is automatically activated.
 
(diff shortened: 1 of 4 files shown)


clintonskitson:
This satisfies issue https://github.com/emccode/rexray/issues/171.

Work to be done including general docs, and covered/not covered scenarios.

Issue #169, opened 2015-10-28
Create volumes with /data directory inside of volume


clintonskitson:
In order to provide applications with empty mount paths and ensure application compatibility, it would be nice to create directories as `/data` on the volume, and pass this off to Docker.

This setting should be able to be set globally through volume specific level.

clintonskitson:
We ran into this problem where the `lost+found` directory was present and the Docker `MySQL` container was detecting this through a custom script and not creating the data structure it needed because it base `/var/lib/mysql` directory as not empty.

Further details to come here.

PR #215, opened 2015-12-15
Added sub-path for volume mounts and permission setting

clintonskitson:
This commit introduces a new setting that ensures that following
the typical mount, that a new directory is created for data
with a default value of /data.  The field is configurable to
ensure compatibility with old versions and future.  The default
permissions on this directory are set to 0700.


clintonskitson:
Satisfies https://github.com/emccode/rexray/issues/169

clintonskitson:
When testing the base MySQL and Postgres containers supplied by Docker, the existing functionality was causing them to fail.

PR #216, opened 2015-12-15
Fixed XIO bug for pre-emption and volume discovery

clintonskitson:
The cautionary unmount was is now performed earlier so that
prior to attaching a device, the mount table is clear.


clintonskitson:
This PR fixes an issue found in 0.3.1-RC1 with XtremIO when enabling pre-emption.

clintonskitson:
The fix was tested to work across all drivers.

PR #228, opened 2015-12-17
Feature/doc updates

clintonskitson:
This PR is a consolidation of doc updates that are intended for the 0.3.1 release.  There is still work that needs to be done here for each commit but it is a good start.  Not intending to merge this as is.

### DIFF
--- a/.docs/index.md
+++ b/.docs/index.md
@@ -121,7 +121,7 @@ $ docker run -d --volume-driver=rexray -v hellopersistence:/mystore \
   --name temp01 busybox
 $ docker exec temp01 "touch /mystore/myfile"
 $ docker exec temp01 "ls /mystore"
-$ docker stop temp01 && docker rm temp01
+$ docker rm -f temp01
 ```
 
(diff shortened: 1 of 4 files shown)


akutz:
Hi @clintonskitson,

Are you satisfied with this PR's content? You said you don't wish to merge as is, but I wasn't sure if the cause was lack of content or some other reason. If it's to clean up the commits I can do that when I have a moment. Just let me know.

clintonskitson:
Some of the commits in here are unfinished.  I will submit other PRs that would include discrete section updates.

clintonskitson:
I am good with these doc updates.

akutz:
Hi @kacole2, @dvonthenen, @drumulonimbus,

Your doc updates look great, but before we can accept them we need you to limit your lines to 80 chars per all of the other markdown files in the docs section. This is a standard to which we adhere to make the sources more readable and to make writers consider brevity always. @kacole2, you do a fantastic job with brevity, so thank you. It's just that we want hard line breaks to boot. Thanks all.

clintonskitson:
Looks good to me now.

Issue #221, opened 2015-12-16
Error details are being squashed


clintonskitson:
Just noticed that the error details are being squashed depending on the type of return.

```
	medium, err := d.virtualbox.CreateMedium("vmdk", d.storageLocation(name), size)
	if err != nil {
		fmt.Println(err)
		return nil, err
	}
	return medium, err
```

```
FATA[0000] error creating new volume                     inner=Method 'IVirtualBox_createHardDisk' not implemented: method name or namespace not recognized provider=virtualbox size=100 snapshotID= volumeID= volumeName=tt.vmdk
```

```
return d.virtualbox.CreateMedium("vmdk", d.storageLocation(name), size)
```

```
INFO[0000] storage driver initialized                    provider=virtualbox
FATA[0000] error creating new volume      
```

akutz:
Why the `fmt.Println`? Also, not sure what you mean on type of return. Do you mean type of log statement?

clintonskitson:
`return d.virtualbox.CreateMedium("vmdk", d.storageLocation(name), size)` this line seems to squash the error compared to setting it as a `medium,err :=` and then passing those specifically.

akutz:
Hi @clintonskitson,

I imagine it's a go-ism on how the return values are inferred? Not sure. Also, are you sure the two log lines `error creating new volume` are being logged from the same location? Maybe one is using a log function format that does not pass the error object to the log function directly, bypassing goof's logic.

clintonskitson:
I updated the output there to reflect what I believe I was seeing.  I will look at it further to be sure.

akutz:
Hi @clintonskitson,

Could you link to some of the sources so it can be traced? Looking at it in isolation makes it difficult to gauge what might possibly be occurring. What appears to be happening is two different places appear to be logging the `err` object, one logging only the `err`'s message and one logging the object using `goof` and `golf`.

clintonskitson:
Looks like it's related to Travis not using the right Sirupsen package when compiling.

```
Binary: /usr/bin/rexray
SemVer: 0.3.1-rc2
OsArch: Linux-x86_64
Branch: v0.3.1-rc2
Commit: b2ebf2c351952eeaa3996398684883ef61c6d09a
Formed: Wed, 16 Dec 2015 23:31:17 UTC

[root@localhost vagrant]# rexray volume mount --volumename=abc3.vmdk
FATA[0000] error attaching volume                       
[root@localhost vagrant]# ./rexray version
[...]

PR #232, opened 2015-12-20
Removed travis caching for Go packages

clintonskitson:
The caching of Go packages is causing inconsistency between the
builds that are supposed to occur from scratch.

This fix addresses issues around the error details being squashed
since the wrong logrus package was being used.

https://github.com/emccode/rexray/issues/221

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,5 +51,4 @@ deploy:
 cache:
   apt: true
   directories:
-    - $GOPATH/pkg
     - $HOME/.opt


akutz:
Hi @clintonskitson,

I'd rather not do this. Instead you can tell Travis to remove the cache so it's re-cached. We should only have to do this when there are changes to glide or some such thing. It does save us quite a bit of time. Do you mind removing the cache via Travis so it rebuilds to see if that works at all?

clintonskitson:
Could but it doesn't feel like a long term solution.  If we can't be
assured that the build process is producing repeatable artifacts then it
seems like a problem.

The way I repeated this locally was to clear out local pkg Dir then build a
really old commit.  From there built from new commit and saw problem
locally.  Cleared pkg and tried to rebuild from new and it was fixed.

It seems like the easiest short term thing for consistency would be to stop
caching for now.  I am open for both solutions but it feels like Go has
some more gremlins to tackle before we can rely on an existing pkg not
screwing up a build?

akutz:
Hi @clintonskitson,

I'm not sure I agree if you'll bear with me on my counter-position. We haven't seen this issue before, and it only occurred due to the way we vendor forked support packages. I don't know that we do that *that* often. 

Still, as you said, we *would* like assurances that these things work as they should for blessed builds.

What's interesting about your case is that the package in question was vendored. I'd be very interested to know if glide was not pulling the correct fork due to one version already existing. I'm not sure why the pkg directory should influence that except to say that go did not find it necessary to rebuild things. And yet if the sources had changed, which they apparently did not, it should have rebuilt it.

I'd like to table this until it happens again, or if you can reproduce it and let me examine the logs I'd like to do that. I do not want the builds to become obscenely long due to losing the cache, but nor do I want what you described -- non-deterministic results.

Since no logs were captured that showed glide's output, I cannot say whether or not glide was pulling the correct repo. I'd like to be able to see that first before we implement this change. And again, this is the first time this has happened that the build has broken, so I'm not sure failing to implement this change now will be immediately detrimental.

clintonskitson:
Ok

clintonskitson:
I tried removing the cache but it looks like the build times are consistent with prior.  I also rebuilt the build but the resulting artifact still did not properly show errors.

PR #234, opened 2015-12-21
Promoted documentation on README

clintonskitson:
The documentation is now promoted in the README.  There is also
now information about reviewing stable docs at RTD or for a
specific commit.


akutz:
Hi @clintonskitson,

Please wrap markdown at 80 chars where possible. Thanks!

PR #242, opened 2015-12-28
Updated openstack options for reauth capability

clintonskitson:
The reauthenticate flag is now set by default for openstack
to handle situations where keystone has expired an existing
token.

### DIFF
--- a/drivers/storage/openstack/openstack.go
+++ b/drivers/storage/openstack/openstack.go
@@ -184,6 +184,7 @@ func (d *driver) getAuthOptions() gophercloud.AuthOptions {
 		TenantName:       d.tenantName(),
 		DomainID:         d.domainID(),
 		DomainName:       d.domainName(),
+		AllowReauth:      true,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


clintonskitson:
Testing this further today.

akutz:
Hi @clintonskitson,

For some reason I thought I pushed 0.3.1 and didn't. Stupid me. I'm handling right now, and saw this. Do you want to wait on this?

clintonskitson:
Yeah I want to include this so let's wait until I can test it a bit more.
I should be able to get to it later today.

heww:
I'm so sorry that only allowed reauth here does not work because there is a bug of reauth in gophercloud.

First, according rackspace/gophercloud we should close old connection(unauthorized) to avoid memory leak, see more https://github.com/rackspace/gophercloud/issues/504

Second, after reauth in Request method of gophercloud's ProviderClient, the Request will call itself to continue the API request (https://github.com/rackspace/gophercloud/blob/master/provider_client.go#L196) . It should return resp, nil directly if no err happend, otherwise a err "http invalid read on closed request body" will raised.

So, I think that gophercloud used by rexray which is https://github.com/clintonskitson/gophercloud.git shoud fix the problems mentioned above.

reauth

clintonskitson:
Thanks for looking at this @heww.  I have included updates per your comments and tested.

PR #256, opened 2016-01-14
add Ansible info to automation docs

codenrhoden:
Just put in a simple example.  I can link to the source or Galaxy page if that's better, as it lists more complete examples.  But I figure this part is just to let people know it exists.

### DIFF
--- a/.docs/user-guide/automation.md
+++ b/.docs/user-guide/automation.md
@@ -14,7 +14,32 @@ orchestration tools below.  Optionally, Docker is also listed in some examples.
 
 <br>
 ## Ansible
-ToDo
+With Ansible, installing the latest REX-Ray binaries can be accomplished by
+including the `codenrhoden.rexray` role from Ansible Galaxy.  The role accepts
+all the necessary variables to properly fill out your `config.yml` file.
+
+Install the role from Galaxy:
+
+```shell
+ansible-galaxy install codenrhoden.rexray
+```
+
+Example playbook for installing REX-Ray on GCE Docker hosts:
+
+```yaml
+- hosts: gce_docker_hosts
+  roles:
+  - { role: codenrhoden.rexray,
+      rexray_service: true,
+      rexray_storage_drivers: [gce],
+      rexray_gce_keyfile: "/opt/gce_keyfile" }
+```
+
+Run the playbook:
+
+```shell
+ansible-playbook -i <inventory> playbook.yml
+```
 
 <br>
 ## AWS CloudFormation


codenrhoden:
Looks like this needs an update.  @clintonskitson would like the repo moved to emccode, so the paths reference here will change.

akutz:
Hi @codenrhoden,

I also plan on moving the automation docs to a different location as I'm cleaning up doc changes made in my absence.

.docs/user-guide/automation.md
```diff
@@ -14,7 +14,32 @@ orchestration tools below.  Optionally, Docker is also listed in some examples.
...
+```shell
+ansible-galaxy install codenrhoden.rexray
+```
+
+Example playbook for installing REX-Ray on GCE Docker hosts:
+
+```yaml
+- hosts: gce_docker_hosts
+  roles:
+  - { role: codenrhoden.rexray,
```
akutz: Hi @codenrhoden,

Is this valid YAML? I've never seen YAML with a leading and ending curly bracket. Is this supposed to be a map? If so, isn't a map in YAML just another level of key/value pairs? You know what, never mind, I just found that's a way to [represent a mapping](http://symfony.com/doc/current/components/yaml/yaml_format.html).

.docs/user-guide/automation.md
```diff
@@ -14,7 +14,32 @@ orchestration tools below.  Optionally, Docker is also listed in some examples.
 
 <br>
 ## Ansible
-ToDo
+With Ansible, installing the latest REX-Ray binaries can be accomplished by
+including the `codenrhoden.rexray` role from Ansible Galaxy.  The role accepts
+all the necessary variables to properly fill out your `config.yml` file.
```
akutz: Hi @codenrhoden,

The `config.yml` file to which you're referring -- is this the REX-Ray config file? If so, I'm not sure the `hosts` and `roles` keys should be at the root in your example below. Unless this is required by the driver, I'd prefer it if you refactor things so that you're not using the root namespace. Maybe shuffle them under a namespace called `ansil`?

.docs/user-guide/automation.md
codenrhoden: Cool - yeah, it's valid.  And it can be a bit weird, because this YAML is consumed by Ansible, and they've made a few extensions to what they consider valid formats.  They take true YAML, but also some hybrid formats to make their playbooks look "cleaner" in their opinion.

.docs/user-guide/automation.md
```diff
@@ -14,7 +14,32 @@ orchestration tools below.  Optionally, Docker is also listed in some examples.
...
+
+Install the role from Galaxy:
+
+```shell
+ansible-galaxy install codenrhoden.rexray
```
akutz: Hi @codenrhoden,

I will add this to the developer docs as part of the doc style guide, but please add a leading `$` when representing individual shell commands. It helps demarcate commands. Thank you.

.docs/user-guide/automation.md
codenrhoden: @akutz, not sure I follow you here.  This snippet is just an example of what would go in an Ansible playbook.  The `config.yml` file I mention is the REX-Ray YAML file, but the code is what goes to Ansible.  Ansible _generates_ the REX-Ray YAML file from what you give it. Does that help?

As a full example, here is a complete playbook (playbook is just a top-level YAML file you feed to ansible) that I used during testing:

```
travis$ cat rexrayrole.yml
---

- hosts: all
  sudo: yes
  roles:
  - { role: ansible-role-rexray,
      rexray_service: false,
      rexray_storage_drivers: [virtualbox],
      rexray_vbox_endpoint: "http://10.0.2.2:18083",
      rexray_vbox_volume_path: "/Users/travis/VirtualBox VMs/Volumes"
```

You'll see it is at the root -- and is not actually related to the `config.yml` file.

akutz: Hi @codenrhoden,

All, see, you presume that I know what the hell about which I'm talking. Or that I think before I speak. I understand now. This looks fine then. I thought the YAML you listed was going **into** the REX-Ray config file, and I was concerned about the `hosts`, `sudo`, and `roles` keys at the root of the config. I understand better now what is occurring. Thank you very much for taking the time to explain it for me.

codenrhoden: No problem!  Thanks for the merge.

I'll update the other piece to add the `$` and change the path to the role in a second PR once the repo move has happened.

akutz: Well crap, I forgot I asked you to do that. It's that kind of day...

Yeah, that stuff can be added / modified later. Thanks!

PR #267, opened 2016-01-25
Updated ScaleIO driver for dynamic storage pools

clintonskitson:
This commit introduces an update that allows multiple storage
pools to be used outside of configuring them in the global
configuration.  The volumeType parameter represents the storage
pool name

### DIFF
--- a/drivers/storage/scaleio/scaleio.go
+++ b/drivers/storage/scaleio/scaleio.go
@@ -199,11 +199,39 @@ func (d *driver) GetVolumeMapping() ([]*core.BlockDevice, error) {
 
 func (d *driver) getVolume(
 	volumeID, volumeName string, getSnapshots bool) ([]*types.Volume, error) {
-	volumes, err := d.storagePool.GetVolume("", volumeID, "", volumeName, getSnapshots)
+	volumes, err := d.client.GetVolume("", volumeID, "", volumeName, getSnapshots)
 	if err != nil {
 		return nil, err
 	}
 	return volumes, nil
+}
+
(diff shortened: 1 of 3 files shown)


clintonskitson:
Testing this further.

clintonskitson:
This is ready to merge upon successful tests.

drivers/storage/scaleio/scaleio.go
```diff
@@ -686,6 +791,14 @@ func (d *driver) storagePoolName() string {
 	return d.r.Config.GetString("scaleio.storagePoolName")
 }
 
+func (d *driver) thinOrThick() string {
+	thinOrThick := d.r.Config.GetString("scaleio.thinOrThick")
+	if thinOrThick == "" {
+		return "ThinProvisioned"
```
akutz: Hi @clintonskitson,

I'd suggest this as an alternative:

```go
type ScaleioProvisionType int

const (
    Thin ScaleioProvisionType = iota
    Thick
)
```

This avoids free-text in many places, other than parsing it from the CLI.

clintonskitson:
Hmm.. having a hard time consuming this one as to the benefit.  Seems like an extra layer of conversion considering it is leaving as a string into `goscaleio`.

akutz:
Hi @clintonskitson,

Eh, you're probably right.

akutz:
Hi @clintonskitson,

This LGTM. It just needs to be merged/rebased with/on master so I can merge it.

clintonskitson:
@akutz Rebased.

PR #272, opened 2016-02-01
Updated unit file for restart

clintonskitson:
This commit introduces the ability for SystemD to restart
the service when it fails.  In the case of a system start
where dependencies for driver starts cannot be satisfied,
the service would previously just fail to start.  Setting
the restart enables retries until the storage driver
can be initialized.

### DIFF
--- a/rexray/cli/installer.go
+++ b/rexray/cli/installer.go
@@ -327,6 +327,8 @@ Description=rexray
 Before=docker.service
 
 [Service]
+Restart=always
+RestartSec=5
 EnvironmentFile={{.EnvFile}}
 ExecStart={{.RexrayBin}} start -f
 ExecReload=/bin/kill -HUP $MAINPID


akutz:
There is a reason this was not implemented originally. If REX-Ray is misconfigured then the service will simply enter an infinite restart loop. Per the [docs](http://www.freedesktop.org/software/systemd/man/systemd.service.html), we'd likely want to set `Restart=on-success` instead of `always`.

> In this context, a clean exit means an exit code of `0`, or one of the signals `SIGHUP`, `SIGINT`, `SIGTERM` or `SIGPIPE`, and additionally, exit statuses and signals specified in `SuccessExitStatus=`.

To that end, if there are occasions that the service is failing to start at boot due to boot order, then the correct approach involves:

1. Modify the unit file to depend on some other service that ensures the storage connection is up before REX-Ray starts.

2. Modify the REX-Ray binary to always exit with one or more specific process exit codes indicating that the failure was due to the inability to access a remote resource. These exit codes can be added to the `SuccessExitStatus` property in the unit file to indicate that a restart attempt should still occur as the exit type was valid.

clintonskitson:
@akutz I believe the proper catch all would be to attempt a restart regardless.  I would be against this if it would prompt a DOS type event, but the timeout is set at a low level.  I believe requiring that a perfect configuration of dependencies happens would be an explicit change that a user would make versus our default behavior.

akutz:
Hi @clintonskitson,

Hmm, I think I still disagree.

1. This definitely causes an infinite loop of restart attempts that will never succeed if the file is misconfigured.
2. The timeout is 5s. That's an attempt to initialize all configured storage drivers every 5s. Some *will* succeed. Maybe this isn't a DOS, but it could be enough to have a remote platform rate limit us or flag us for odd behavior.

Regarding the endless loop of restart attempts, the following properties should be added to the unit file:

```
StartLimitInterval=60s
StartLimitBurst=5s
```

### StartLimitInterval=, StartLimitBurst=

Configure service start rate limiting. By default, services which are started more than `5` times within `10` seconds are not permitted to start any more times until the `10` second interval ends. With these two options, this rate limiting may be modified. 

Use `StartLimitInterval=` to configure the checking interval (defaults to `DefaultStartLimitInterval=` in manager configuration file, set to `0` to disable any kind of rate limiting). Use `StartLimitBurst=` to configure how many starts per interval are allowed (defaults to `DefaultStartLimitBurst=` in manager configuration file). 

These configuration options are particularly useful in conjunction with `Restart=`; however, they apply to all kinds of starts (including manual), not just those triggered by the `Restart=` logic. Note that units which are configured for `Restart=` and which reach the start limit are not attempted to be restarted anymore; however, they may still be restarted manually at a later point, from which point on, the restart logic is again activated. 

Note that systemctl reset-failed will cause the restart rate counter for a service to be flushed, which is useful if the administrator wants to manually start a service and the start limit interferes with that.

clintonskitson:
@akutz I believe the infinite restarts is fine by default.  This should be super simple out of the gate.  Having a rate limiting may still introduce the problem that this thread was based on. 

There could be documentation that points people to read up on the proper options if they want to customize the restart capabilities.

akutz:
Hi @clintonskitson,

You'll need to use a managerial override on this. If you do, then please update the documentation to note that the REX-Ray service will fail into an infinite restart loop if the configuration file is in error, a remote resource is unavailable, etc. My dissenting opinion is that we should err on the side of caution and fail fast, providing documentation that explains to users how to shoot themselves in the foot, but not loading the firearm and cocking the firing pin for them.

clintonskitson:
@akutz Let's revisit this one at a later point.  I see both sides of the argument, but it is a minor change for someone who has special needs.  Let's see how it pans out re future issues.

akutz:
Hi @clintonskitson,

That's fine, but I think you deflated your own argument. If it's a minor change for someone with special needs, then they should just deploy with custom unit file instead of the stock. Needs of the many and all that. In this case this PR is for the few. Or rather, for a single person as it sounds.

PR #274, opened 2016-02-02
Added support for Volume Driver with Docker 1.10

clintonskitson:
This commit introduces new functionality that is exposed to the
Docker Volume Driver in Engine 1.10.  Calls for Get/List are
now implemented to enable the engine to rely on volume driver
to return list of existing volumes.  A list locally is no longer
stored which means the array is the single source of truth
for available volumes.

### DIFF
--- a/.docs/index.md
+++ b/.docs/index.md
@@ -104,7 +104,7 @@ $ rexray start
 ```
 
 ### Example of REX-Ray with Docker
-Docker 1.9.1+ is recommended to use REX-Ray as a
+Docker 1.10+ is recommended to use REX-Ray as a
 [Docker Volume Driver Plugin](https://docs.docker.com/extend/plugins_volume/).
 
 The following example uses two Amazon EC2 Virtual Machines, `EC2a` and `EC2b`,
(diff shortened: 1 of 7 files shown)


clintonskitson:
This relates to https://github.com/docker/docker/commit/d3eca4451d264aac564594fe46b8c097bd85a5cc

akutz:
Hi @clintonskitson,

Please increase the code coverage if possible so that the number doesn't drop any lower. It's dropped nearly 6% from where it was December 1st.

Also, is this the type of thing that should be added during the RC process? I'd lean towards no, and defer it to 0.3.3 in order to ensure that the RC process avoids new functionality and instead focuses on tidying things up as well as minor bug fixes.

clintonskitson:
@akutz Code coverage updated.  

Yes it would be great to be able to lock in feature work pre-RC and then only iterate on fixes for the release.  This one puts us in a tough position because the change from Docker's perspective "Existing plugins need to make use of these new APIs to satisfy users' expectation" for 1.10.  Unfortunately the API is not perfectly versioned today so getting aligned to the API sooner is a good idea.  On our end the changes were net new method calls and didn't change existing functionality.

I will test further across different revs and get back to the thread here.

akutz:
Hi @clintonskitson,

It's still not able to merge due to lack of coverage. You can check this yourself via the GitHub service checks section above:

image

If you click the [Details](https://coveralls.io/builds/4952677) link next to the fails coverage indicator, Coveralls explains which file lost coverage and how to fix it.

clintonskitson:
@akutz Found the missing pieces.  Coverage updated.

PR #284, opened 2016-02-04
Fixes scope problems in docker volume driver

clintonskitson:
The introduction of gofig with scoping caused the values that
were customizable for the docker driver to not work anymore.  This
commit brings that functionality where scoped docker parameters
can be included now in modules and generally.

### DIFF
--- a/drivers/volume/docker/volume.go
+++ b/drivers/volume/docker/volume.go
@@ -39,9 +39,38 @@ func newDriver() core.Driver {
 	return &driver{}
 }
 
+func ef() goof.Fields {
+	return goof.Fields{
+		"provider": providerName,
+	}
+}
+
+func eff(fields goof.Fields) map[string]interface{} {
+	errFields := map[string]interface{}{
+		"provider": providerName,
+	}
+	if fields != nil {
+		for k, v := range fields {
+			errFields[k] = v
+		}
+	}
+	return errFields
+}
+
 func (d *driver) Init(r *core.RexRay) error {
 	d.r = r
-	log.WithField("provider", providerName).Info("volume driver initialized")
+
+	fields := eff(map[string]interface{}{
+		"volumeType":       d.volumeType(),
+		"iops":             d.iops(),
+		"size":             d.size(),
+		"availabilityZone": d.availabilityZone(),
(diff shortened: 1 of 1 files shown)


drivers/volume/docker/volume.go
```diff
@@ -804,6 +832,22 @@ func (d *driver) volumeRootPath() string {
 	return d.r.Config.GetString("linux.volume.rootPath")
 }
 
+func (d *driver) volumeType() string {
+	return d.r.Config.GetString("docker.volumeType")
+}
+
+func (d *driver) iops() string {
```
akutz: Hi @clintonskitson,

Please use the `GetString` function directly instead of getting it as an int and casting it to a string. The underlying code should handle the casting for you. I think.

drivers/volume/docker/volume.go
```diff
@@ -804,6 +832,22 @@ func (d *driver) volumeRootPath() string {
...
+	return strconv.Itoa(d.r.Config.GetInt("docker.iops"))
+}
+
+func (d *driver) size() string {
+	return strconv.Itoa(d.r.Config.GetInt("docker.size"))
```
akutz: Hi @clintonskitson,

Please use the `GetString` function directly instead of getting it as an int and casting it to a string. The underlying code should handle the casting for you. I think.

akutz:
Hi @clintonskitson,

To be fair, it wasn't scoping that was the issue. The Docker code was simply not using Gofig *at all* -- scoping or otherwise. If the env vars were still set then it would have still worked.

drivers/volume/docker/volume.go
clintonskitson: Yup, my fault.

clintonskitson:
Roger that.

clintonskitson:
@akutz Updated per your comments.

akutz:
Hi @clintonskitson,

Thanks. I should have updated the driver. That's not in question. I just wanted to have it pointed out that it was my lack of updating the driver at all, not the introduction of scoping, that broke things. I don't want any questions of stability regarding scoping unless they do point to real issues.

clintonskitson:
Still testing this.

clintonskitson:
@akutz LGTM here.

Issue #295, opened 2016-02-12
Reporting module name in log


clintonskitson:
With the introduction of the modules, it has introduced a necessary context that should be tracked in the logs.  Right now, when requests come in you can't differentiate with module or storage provider is responding to the request.

This log was generated from a single `docker` instance talking with three volume endpoints that are all pointing to the same storage platform.

I believe this is part of some of the 0.4 changes that @akutz has planned re context, but wanted to make sure we track it.

```
time="2016-02-12T15:50:26Z" level=info msg="getting volume" driverName=docker volumeName=newvol2001 
time="2016-02-12T15:50:26Z" level=info msg="getting volume" driverName=docker volumeName=newvol2001 
time="2016-02-12T15:50:32Z" level=info msg="getting volume" driverName=docker volumeName=newvol2001
```

akutz:
Hi @clintonskitson,

I opened a PR where I implemented a short-term fix for this.

clintonskitson:
https://github.com/emccode/rexray/pull/296

PR #296, opened 2016-02-12
Add REX-Ray Context

akutz:
This patch adds the field `Context` to the REX-Ray struct, enabling modules to set the context for a REX-Ray runtime for future logging.

This PR handles issue #295.

### DIFF
--- a/core/drivers_os.go
+++ b/core/drivers_os.go
@@ -86,9 +86,10 @@ func (r *odm) GetMounts(
 	deviceName, mountPoint string) (MountInfoArray, error) {
 	for _, d := range r.drivers {
 		log.WithFields(log.Fields{
+			"moduleName": r.rexray.Context,
 			"deviceName": deviceName,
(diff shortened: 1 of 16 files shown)


clintonskitson:
Ok, great.  That helps at the high level to see where the calls are coming from.

```
time="2016-02-12T22:31:08Z" level=info msg="listing volumes" driverName=docker moduleName=vb2 
time="2016-02-12T22:31:08Z" level=info msg="listing volumes" driverName=docker moduleName=vb3 
```

akutz:
Hi @clintonskitson,

It can be referenced anywhere that has a handle to the RexRay instance. All drivers should store a reference to it...

akutz:
Ultimately I want to extend logrus to handle this more natively, but I've been waiting for their big refactor and point release as they're deferring all PRs until then. However, the absence of Threads and thus ThreadLocals in Go makes logging contexts difficult. That's why Google recommends passing the Context object everywhere.

PR #304, opened 2016-02-18
Added limitation for volume name for SIO

clintonskitson:
This commit introduces a new volume name limit that pertains
specifically to the SIO storage platform. The limit is 31 chars
enforced by SIO. A bug was introduced by Docker 1.10 since it
started creating anonymous volumes when no volumes were specified
which defaulted to 32 characters.

### DIFF
--- a/drivers/storage/scaleio/scaleio.go
+++ b/drivers/storage/scaleio/scaleio.go
@@ -16,6 +16,7 @@ import (
 )
 
 const providerName = "ScaleIO"
+const cc = 31
 
 // The ScaleIO storage driver.
 type driver struct {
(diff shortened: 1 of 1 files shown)


akutz:
Any reason not to make CC configurable or a pkg constant? Or even at the VDM level?

clintonskitson:
I was thinking about this for a bit. I would pursue the VDM route, but I have not run into the situation with any other storage driver. For now I have made it a constant instead, and kept it internal to the driver. I am open to revisiting later to graduate the parameter if we feel it is necessary for other drivers as well or should be configurable
.

akutz:
It's not currently a constant; nor pkg level. Want to correct that? Currently it's a variable that allocates memory on each func call.

akutz:
Nevermind, when you said "for now" you meant you went back and corrected it. Cool.

PR #306, opened 2016-02-19
Docker cache volumes for path operations

clintonskitson:
This introduces a volume caching feature for Path lookups. Any Path operations will only result in a future lookup if a volume 1) is mounted when rr started 2) mount operation occurs on it. This addresses a Docker 1.10.1 problem issue where Path requests are made across every volume that has not Mountpoint listed from List requests.

This should be rebased after https://github.com/emccode/rexray/pull/296 is merged.  No conflicts should exist at that point.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -744,7 +744,23 @@ sharing volumes, it is recommended that you reset the service along with the
 accompanying container runtime (if this setting is false) to ensure they are
 synchronized.  
 
-#### Volume Path (0.3.1)
+#### Volume Path Disable Cache (0.3.2)
+In order to minimize the impact to return `Path` requests, a caching
+capability has been introduced by default. A `List` request will cause the
+returned volumes and paths to be evaluated and those with active mounts are
+recorded. Subsequent `Path` requests for volumes that have no recorded mounts
+will not result in active path lookups. Once the mount counter is initialized or
+a `List` operation occurs where a mount is recorded, the volume will be looked
+up for future `Path` operations.
+
+```yaml
+rexray:
+  volume:
+    path:
+      disableCache: true
+```
+
+#### Volume Root Path (0.3.1)
 When volumes are mounted there can be an additional path that is specified to
 be created and passed as the valid mount point.  This is required for certain
 applications that do not want to place data from the root of a mount point.  
--- a/core/drivers_volume.go
+++ b/core/drivers_volume.go
@@ -88,6 +88,14 @@ func (r *vdm) Init(rexray *RexRay) error {
 		return errors.ErrNoVolumeDrivers
 	}
 	r.mapUsedCount = make(map[string]*int)
+
+	log.WithField("pathCache", r.pathCache()).
+		Debug("checking volume path cache setting")
+
+	if r.pathCache() {
+		_, _ = r.List()
+	}
+
 	return nil
 }
 
@@ -186,12 +194,19 @@ func (r *vdm) countRelease(volumeName string) {
 }
 
 func (r *vdm) countExists(volumeName string) bool {
-	_, exists := r.mapUsedCount[volumeName]
-	log.WithFields(log.Fields{
+	fields := log.Fields{
 		"moduleName": r.rexray.Context,
 		"volumeName": volumeName,
-		"exists":     exists,
-	}).Info("status of count")
+	}
+
+	var c *int
+	var exists bool
+	if c, exists = r.mapUsedCount[volumeName]; exists {
+		fields["count"] = *c
+	}
+	fields["exists"] = exists
+
+	log.WithFields(fields).Info("status of count")
 
 	return exists
 }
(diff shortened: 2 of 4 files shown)


clintonskitson:
@akutz Not sure how to satisfy the testing here with the unexported methods.

clintonskitson:
This other important note here is that in order for this to work, it introduces a List() operation when the volume driver starts. This populates a count for any volumes that have existing mounted paths. This value is used to determine whether the future Path() operations would require any lookups. The feature is globally able to be disabled with `volume.path.disablecache`.

An outstanding note is that since it does this with the `vdm` Init, a List() happens before each CLI command even though the cache isn't relevant for those commands. I couldn't figure out how best to only restrict this to the daemonized modules. A solution would be to have a minimum of 2 modules configured `default-docker` that disables the cache and is used for CLI, then `custom1` that is used as the docker endpoint which would include the cache.

akutz:
Hi @clintonskitson,

I'm going to need to spend some time reviewing the VDM code to figure out the best path forward.

akutz:
Hi @clintonskitson,

Also, please rebase this branch ASAP. It's got commits on it that are no longer in existence as they've been merged after having been rebased, and will have different commit IDs than what are on this branch. Thanks.

akutz:
Hi @clintonskitson,

Would you please provide clarity to the following remarks:

> An outstanding note is that since it does this with the vdm Init, a List() happens before each CLI command even though the cache isn't relevant for those commands. 

Is the cache relevant to *any* CLI command, or only for modules? And if so, only for *some* modules? Because if that's the case, why even invoke it inside the VDM and not in the Docker module instead? It seems maybe too aggressive to implement a Docker fix at a global scope.

You're correct regarding the side-effect -- it will prevent this PR from being merged as is. The most prudent and simplest path forward is to reverse the meaning of the new switches/config options. Please change `rexray.volume.path.disableCache` to `rexray.volume.path.enableCache`, and then in the Docker **Daemon Module** (not the Docker Volume driver, but the **Daemon Module**), prior to instantiating a new REX-Ray instance, create a new Config instance and enable the cache.

For example, on [line 59 of daemon/module/docker/volumedriver/voldriver.go](https://github.com/emccode/rexray/blob/master/daemon/module/docker/volumedriver/voldriver.go#L59), change this:

```
r := core.New(c.Config)
```

So that the when creating a new Docker Daemon module, the supplied Config instance is copied (so that the next step doesn't affect other modules), and then the `rexray.volume.path.enableCache` setting is flipped to `true`:

```
cc, err := c.Config.Copy()
if err != nil {
    return nil, err
}
cc.Set("rexray.volume.path.enableCache", true)
r := core.New(cc)
```

This way the existing behavior of REX-Ray is preserved and only the daemon modules requiring caching will get it. If this is not acceptable then what we should do is to alter how a REX-Ray instance is constructed and enable passing an additional parameter based on an enum that describes the purpose of this instance:

```
type RunAsTypes int
const (
    RunAsCLI RunAsTypes = iota
    RunAsService
)
```

Then a REX-Ray instance can store a field of type `RunAsTypes` and subsequent logic can use that to detect if there's something that should or should not occur based on how REX-Ray is being utilized.

clintonskitson:
@akutz Thanks for the review.  I have re-based and modified based on your comments. I did introduce the configurable `path.disableCache` option. The objective was to ensure the cache was enabled by default for the docker modules. Open to alternatives since I use `path.cache` as an internal param and `path.disableCache` as the configurable item. I believe the configRegistration would simplify this, but wasn't sure if it could be introduced in the right spot to make it useful here for default values.

akutz:
Hi @clintonskitson,

First of all, thank you for rebasing the PR as I requested. It makes the review a lot easier :) Secondly, I'm not sure that I successfully communicated the request for changing the name of the switch. The fact is, caching should *not* be enabled by default for *all* volume drivers since, as you said, it's Docker specific. Additionally, since my question regarding whether caching is applicable in CLI mode was left unanswered, I have to restate one of the reasons for defaulting to **no** cache by default -- this prevents REX-Ray from listing all paths via the VDM every time a CLI command is invoked. If instead the suggested fix is used, and the config option defaults to a disabled cache, requiring explicit enabling, then we can **always** and **only** enable it for the Docker Daemon Module so that the Docker Volume Driver under such module has caching. All other instances of any Volume driver, including the Docker Volume Driver, will default to **no cache**, including the CLI. 

So please clarify if:

1. Caching is only needed for REX-Ray as a service/module 
2. Caching is not useful and should not be enabled for the CLI
3. Caching is specific to the Docker Volume Driver **when** operating via the Docker Daemon Module (related to the first point)

If these assumptions/assertions are not accurate, then I've misunderstood the PR and its explanation. As it is it seems the updated PR still does not accommodate the issue that you pointed out regarding the caching forcing a listing of the paths with every CLI command.

clintonskitson:
@akutz I believe your understanding is correct.

My intent here is to ensure that by default, the caching feature is enabled for any daemonized volume module. It is in response to the Docker 1.10 API, but it was introduced in a way that would have benefit to any daemonized volume module. I am not sure why someone would disable the functionality, hence the desire to ensure it is default.

I believe the state of the PR is what you have defined.

clintonskitson:
@akutz Also based on how the parameter is interpreted, it is only actually activated if the daemon module implementer manually evaluates and flips the bit.

https://github.com/clintonskitson/rexray/blob/feature/docker_cache_volumes/daemon/module/docker/volumedriver/voldriver.go#L64-L66

PR #307, opened 2016-02-19
Fix Docker XIO create problem

clintonskitson:
This commit fixes a problem where the `docker volume create` command was not working. The fix to ensure there is no error being sent back in the case of a volume not found.

This should not be merged until it is rebased from https://github.com/emccode/rexray/pull/296

Intentional since the context configuration is set in this PR expecting the merge to occur at core.


akutz:
Please rebase per your note above, then it LGTM

clintonskitson:
@akutz This has been rebased.

PR #309, opened 2016-02-19
Glide 0.9 Resolve Current Fix

akutz:
This patch fixes an issue with Glide 0.9 causing an error when it attempts to resolve dependencies across multiple Go OS/Arch combinations.


akutz:
Hi @clintonskitson,

Make sure you go `go get -u github.com/Masterminds/glide` to upgrade to the latest tip of master for Glide, or else the updated Makefile won't work for you locally.

Issue #294, opened 2016-02-12
Perform YAML validation of config files


clintonskitson:
I have had YAML formatting problems numerous times that resulted in the storage driver not starting correctly.  Troubleshooting was able to narrow it down to invalid characters etc in the YAML files.

It would help the user experience a lot when configuring new and more complex drivers to perform a basic validation of the config.yml file prior to attempting a service start with it.

akutz:
Hi @clintonskitson,

I agree this is a good idea, and I've even looked into it in the past, but Viper has always failed gracefully, ignoring bad content, when loading files. To that end I've just submitted a [PR to Gofig](https://github.com/akutz/gofig/pull/6) that can be used to validate YAML. See the test [TestValidateYAML](https://github.com/akutz/gofig/blob/master/gofig_test.go#L137).

clintonskitson:
Very nice @akutz.  I just did some basic testing with it and it is going to help a lot.  The package looks to be pretty verbose with errors.  Maybe we should add in a brief blurb into the docs in the configuration section about taking the config to yamllint.com if there are errors in their configs.

```
 line 4: found a tab character that violate indentation
```

akutz:
Hi @clintonskitson,

Well, I'm not controlling content of said errors. To what type of verbosity are you referring? Also, I had no intention of necessarily printing out the direct errors. We could always catch the error and print out to the user that they should take the YAML and submit it to yamllint.com themselves. We can't expect to have network access to do it programmatically, but we can let them know to do it on error.

clintonskitson:
Yah exactly.  I was actually just thinking about calling it out with a couple of sentences in the docs themselves.  But an explicit comment to verify it online or with a lint would work as well.

What would be your concern with repeating the errors that YAML returns?

akutz:
I have no such concern; from your comments above it seemed you did regarding the errors' verbosity.

clintonskitson:
Nope, I believe they would be very helpful for people configuring the service.

akutz:
> The package looks to be pretty verbose with errors. Maybe we should add 

The declaration of verbosity followed by a recommendation seems to me to indicate "Because it does this, we should do that as a counter/enhancement."

Just a case of intent and tone lost in text communication :)

clintonskitson:
Yeah, I missed the "I like this" part. =)

clintonskitson:
@akutz Is this something we are comfortable with introducing into 0.3.2?

PR #311, opened 2016-02-21
YAML Validation

akutz:
This patch adds YAML validation to prevent errant characters in the config files from causing hard-to-detect runtime misconfigurations. This is in response to issue #294. Please see the following evidence of the fix. Note in the below config file, the last line should be indented beneath `drivers:`

```
$ cat /etc/rexray/config.yml 
osDrivers:
- mockOSDriver
volumeDrivers:
- mockVolumeDriver
storageDrivers:
- libstorage
libstorage:
  host: "127.0.0.1:1870"
  profiles:
    enabled: true
    client: true
    groups:
    - local=127.0.0.1
  drivers:
- mockStorageDriver
```

Now here is what happens when executing REX-Ray:

```
[0]akutz@pax:rexray$ rexray
rexray: invalid config: /etc/rexray/config.yml

  yaml: line 14: did not find expected key

paste the contents between ---BEGIN--- and ---END---
into http://www.yamllint.com/ to discover the issue

---BEGIN---
osDrivers:
- mockOSDriver
volumeDrivers:
- mockVolumeDriver
storageDrivers:
- libstorage
libstorage:
  host: "127.0.0.1:1870"
  profiles:
    enabled: true
    client: true
    groups:
    - local=127.0.0.1
  drivers:
- mockStorageDriver

---END---
[1]akutz@pax:rexray$ 
```

The fix validates the global and user config files as well as any config path specified on the command line.

### DIFF
--- a/rexray/cli/cli.go
+++ b/rexray/cli/cli.go
@@ -3,10 +3,12 @@ package cli
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/akutz/gofig"
 	glog "github.com/akutz/golf/logrus"
 	"github.com/akutz/gotil"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


clintonskitson:
This tests good by me @akutz.

PR #312, opened 2016-02-23
Updated docker example

clintonskitson:
The index file has been updated for a valid docker example.


akutz:
Which is why I didn't push the RC, because you always one or two more the next day :)

-- 
-a

clintonskitson:
Par for the course I suppose =)

PR #319, opened 2016-02-25
Fix for no logging in 0.3.2-RC3/RC4

akutz:
This patch should fix the issue of no logging in 0.3.2-RC3/RC4 built on Travis. The error was due to the way Glide changed transitive dependency discovery and was no longer using the Logrus package referenced by REX-Ray's referenced Goof package. The correct Logrus package is now explicitly listed in the REX-Ray glide.yaml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+go_import_path: github.com/emccode/rexray
+
 language: go
 
 go:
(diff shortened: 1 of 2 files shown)


akutz:
There is an available Travis-built binary for this PR at https://bintray.com/emccode/rexray/unstable/0.3.2-rc5%2B175. Please test.

akutz:
Tested / Merged

PR #321, opened 2016-02-25
Bugfix for volumetype as mount fs

clintonskitson:
This commit fixes the situation where when specifying a
volumeType as a global parameter for Docker, the mount operation
attempts to use that for the fileSystem. The fsType is now
a configurable item as well.


akutz:
Hi @clintonskitson,

Thank you for doing the correct thing and specifying the release branch for this PR. I was not paying attention (I'm in a waiting area on an iPad) when I merged this, and did not think to check that this PR was going into the release branch.

clintonskitson:
Fixes https://github.com/emccode/rexray/issues/320#issuecomment-188886067

PR #314, opened 2016-02-24
Added ec2 tags to volumes and snapshots

echupriyanov:
Now rexray only lists volumes and snapshot with specific tag/value.
Tag is defined in `drivers/storage/ec2.go` driver:
```
const rexrayTag = "rexraySet»
```
Tag value can be defined in `rexray/config.yml`:
```
aws:
  accessKey: ...
  secretKey: ...
  region: ...
  rexrayTag: rexray
```
Default value is `rexray`.

PR #325, opened 2016-02-26
[experimental] Replace goamz with aws-sdk-go

echupriyanov:
This is a direct rewrite using [aws-sdk-go](https://github.com/aws/aws-sdk-go).
Some minor tweaks:
* `rexray` now picks AWS credentials from all known places (environment,
  ~/.aws/credentials and EC2 instance profile)
* `volume attach` - API call _requires_ presence of instanceID.
  If missing, RexRay nevertheless passes null instance ID to AWS getting an error.
  Now driver's EC2 instance ID is used, if none provided.
* `snapshot copy` - as with `volume attach`, destinationregion is mandatory. And if its omitted,
  AWS returns an error. And as with `volume attach`, driver's EC2 instance Region is used as default

Everything else should work as before

Signed-off-by: Eugene Chupriyanov <tchu@tchu.ru>

### DIFF
--- a/drivers/storage/ec2/ec2.go
+++ b/drivers/storage/ec2/ec2.go
@@ -13,11 +13,16 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/akutz/gofig"
 	"github.com/akutz/goof"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
 
 	"github.com/emccode/rexray/core"
 	"github.com/emccode/rexray/core/errors"
-	"github.com/goamz/goamz/aws"
-	"github.com/goamz/goamz/ec2"
 )
 
 const providerName = "ec2"
@@ -26,6 +31,7 @@ const providerName = "ec2"
 type driver struct {
 	instanceDocument *instanceIdentityDocument
 	ec2Instance      *ec2.EC2
+	ec2creds         *credentials.Credentials
 	r                *core.RexRay
 }
 
@@ -76,18 +82,26 @@ func (d *driver) Init(r *core.RexRay) error {
 		return goof.WithFields(ef(), "error getting instance id doc")
 	}
 
-	auth := aws.Auth{
-		AccessKey: d.accessKey(),
-		SecretKey: d.secretKey(),
-	}
 	region := d.region()
 	if region == "" {
 		region = d.instanceDocument.Region
 	}
-	d.ec2Instance = ec2.New(
-		auth,
-		aws.Regions[region],
-	)
+
+	mySession := session.New()
+
+	d.ec2creds = credentials.NewChainCredentials(
+		[]credentials.Provider{
+			&credentials.StaticProvider{Value: credentials.Value{AccessKeyID: d.accessKey(), SecretAccessKey: d.secretKey()}},
+			&credentials.EnvProvider{},
+			&credentials.SharedCredentialsProvider{},
+			&ec2rolecreds.EC2RoleProvider{
+				Client: ec2metadata.New(mySession),
+			},
+		})
+
+	config := aws.NewConfig().WithCredentials(d.ec2creds).WithRegion(region)
+
+	d.ec2Instance = ec2.New(mySession, config)
 
 	log.WithFields(fields).Info("storage driver initialized")
 
@@ -127,9 +141,9 @@ func (d *driver) GetVolumeMapping() ([]*core.BlockDevice, error) {
 			ProviderName: providerName,
 			InstanceID:   d.instanceDocument.InstanceID,
 			Region:       d.instanceDocument.Region,
-			DeviceName:   blockDevice.DeviceName,
-			VolumeID:     blockDevice.EBS.VolumeId,
-			Status:       blockDevice.EBS.Status,
+			DeviceName:   *blockDevice.DeviceName,
+			VolumeID:     *((*blockDevice.Ebs).VolumeId),
+			Status:       *((*blockDevice.Ebs).Status),
 		}
 		BlockDevices = append(BlockDevices, sdBlockDevice)
 	}
@@ -167,15 +181,14 @@ func getInstanceIdendityDocument() (*instanceIdentityDocument, error) {
 	return &document, nil
 }
 
-func (d *driver) getBlockDevices(instanceID string) ([]ec2.BlockDevice, error) {
+func (d *driver) getBlockDevices(instanceID string) ([]*ec2.InstanceBlockDeviceMapping, error) {
 
 	instance, err := d.getInstance()
 	if err != nil {
-		return []ec2.BlockDevice{}, err
+		return nil, err
 	}
 
-	return instance.BlockDevices, nil
-
+	return instance.BlockDeviceMappings, nil
 }
 
 func getInstanceName(server ec2.Instance) string {
@@ -184,8 +197,8 @@ func getInstanceName(server ec2.Instance) string {
 
 func getTag(server ec2.Instance, key string) string {
 	for _, tag := range server.Tags {
-		if tag.Key == key {
-			return tag.Value
+		if *tag.Key == key {
+			return *tag.Value
 		}
 	}
 	return ""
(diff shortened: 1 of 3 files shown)


akutz:
Hi @echupriyanov,

First of all, this is great work and a truly useful contribution -- thank you. I do have some comments:

1. Please evaluate the formatting applied to the file `glide.yaml` and adjust the changes so that the entire file isn't replaced and re-added. Was it some leading spacing perhaps?
2. Have you verified that the same environment variables are used by the official SDK that GoAmz uses to control and influence the library? For a reference of these environment variables, see the next item.
3. Please update any necessary documentation in `.docs/`, specifically in the user guide section.

Please let me know if you have any questions. Thank you again!

echupriyanov:
1. Regarding `glide.yml` - it gets messed up, when I add a new package with `glide get ...`

echupriyanov:
1. Regarding `glide.yml`. It gets messed up, when I add a new package to project, using `glide get ...`, so to keep commit cleaner, I've manually replaced goamz with aws-sdk-go. Is there a better way to manage glide.yml without need to manually update it?
2. To keep updated version of `rexray` compatible with original, I'm using AWS creds configured for `goamz` (either from `config.yml` or defined in environment) as primary source, and turning to AWS-way of providing creds, if none was found.
3. I've updated docs with new methods of credentials provisioning.

akutz:
Hi @echupriyanov,

I don't ever use `glide get` and only ever manually update the file. If `glide get` screws up the file, please avoid using it. I prefer manual updates to the file anyway. Thanks!

akutz:
Hi @echupriyanov,

Could you please rebase this PR off of the current `master`. I'd like to work towards merging this PR. I'd prefer if you chopped b44817e33c5063cbfa5cd9ba91898d03b86cbd5b this off the PR and rebased the PR off of master instead of the merge. It's cleaner IMO.

echupriyanov:
Sorry for a mess. It was my mistake - I clicked `update branch` button on CI status, instead of mrebasing to newer master. Hope its clear now.

drivers/storage/ec2/ec2.go
```diff
@@ -409,61 +446,60 @@ func (d *driver) CreateVolume(
...
-		SnapshotId: snapshotID,
-		AvailZone:  availabilityZone,
-		VolumeType: volumeType,
-		IOPS:       IOPS,
+	options := &ec2.CreateVolumeInput{
+		Size:             &size,
+		SnapshotId:       &snapshotID,
+		AvailabilityZone: &availabilityZone,
+		VolumeType:       &volumeType,
+		Iops:             &IOPS,
```
jdef: this should probably be nil if IOPS == 0

jdef: we've confirmed that this is a problem; the "zero-value" of the IOPS parameter gets passed in here and results in operation errors when volumetype standard is used

echupriyanov: Thanks for pointing it out. I will check and fix this.

echupriyanov: Should work in updated PR

drivers/storage/ec2/ec2.go
```diff
@@ -409,61 +446,63 @@ func (d *driver) CreateVolume(
...
-		VolumeType: volumeType,
-		IOPS:       IOPS,
+	options := &ec2.CreateVolumeInput{
+		Size:             &size,
+		SnapshotId:       &snapshotID,
+		AvailabilityZone: &availabilityZone,
+		VolumeType:       &volumeType,
+	}
+
+	if volumeType == "io1" {
```
jdef: this seems fragile. probably better off checking for `if IOPS > 0`

echupriyanov: Yes. Let's do it that way.
PR updated

clintonskitson:
@echupriyanov We are looking to merge this in on a separate branch for 0.3.3 as we have already begun development on 0.4.0 with the `master`. Can you create a new PR for this going against our 0.3.3 branch?

clintonskitson:
@echupriyanov I was just running through further test on this branch.  It looks like the create volume from volume functionality isn't working as before.

```
root@ip-172-31-10-250:/home/ubuntu# rexray volume create --volumename=test2 --volumeid=vol-8c146c31
FATA[0001] IncorrectState: Snapshot is in invalid state - pending
	status code: 400, request id: 431154ad-0339-4f8a-8dcd-99286a980647 
```

Check the way that things were occurring previously. I believe we would retry this operation until the operation versus querying to see the state due to the "eventually consistent" nature of the API.

clintonskitson:
@echupriyanov Wondering if the docs here should reference a few details about configuring IAM. ie. create role, assign perms to role, when creating instance associate role. Are there a set of IAM perms that you set for minimum privileges?

echupriyanov:
@clintonskitson thanks for discovering a bug in my rewrite. it should work the same way as with `goamz`, but didn’t, because of slight change in error message format.
So I both fixed error check for pending snapshot, and also added a wait for snapshot creation in `createVolumeCreateSnapshot` as I think its more idiomatic.

echupriyanov:
@clintonskitson I’ve added «minimal» IAM policy to documentation. Some operation there aren’t used right now, but can be used in later version.

jdef:
FWIW re-squashing the PR after every change makes it very hard to observe the individual changes made to fix bugs. If this is the standard for the project, would you mind commenting on the PR where you made the change so that it's easier to review?

akutz:
Hi @jdef,

Where did you see instructions to squash anything? 

Also, my policy on PRs is simply that before a PR is accepted it should be the minimal number of commits. Each commit on a PR should represent an atomic unit of work that can be removed. The value of branching IMO isn't to add anything, but to make it easy to pull back out. A PR with 100 commits may be useful to you, but once it's in the main project, if there's an issue, the entire PR will be reverted, and 100 commits or a single commit matters little.

Still, I looked above and don't see anything regarding squashing your commits.

jdef:
@akutz Appreciate minimizing commits. Comments from @echupriyanov suggest that additional changes have been made very recently. But I don't see separate commits for those. SO .. something was either squashed or "fixup"d. I'm used to squashing commits once the PR has passed review and it about to be merged. Perhaps this PR is already at that point.

akutz:
Hi @jdef,

I'm not sure what happened to be honest. I know @clintonskitson has asked @echupriyanov to close this PR and re-open it against the `release/0.3.3` branch so that we can include this PR in the 0.3.3 release. 

I don't ever rebase/fixup/or change commit IDs on `master`, so at best guess what happened here is that @echupriyanov rebased `echupriyanov:aws-official-sdk` on `emccode:master` and that in the process some commits on `echupriyanov:aws-official-sdk` were lost as part of a fix-up. I know I asked @echupriyanov to do that early on in this PR as the PR had several commits that should have been a single one (atomic functionality) in terms of pre-merge clean-up. However, I don't know if such a thing has happened recently.

What I suspect happen is that, as you said, the PR was at the point of merge but then something shifted or changed -- likely the decision by myself and @clintonskitson to publish a 0.3.3 release to include support for ScaleIO 2.0. When we realized 0.3.3 was on the table, @clintonskitson decided to include this PR as well, and it's likely some commits are now occurring on this PR as a result of that ask.

@clintonskitson, @echupriyanov, care to comment? Am I way off base?

echupriyanov:
Actually, I have 2 active PRs. 
The first one is for adding tags to created volumes/snapshots (https://github.com/emccode/rexray/pull/314). Its based on original `goamz` and @clintonskitson asked me to reopen it basing on release-0.3.3 branch, so I did that.

At the same time @clintonskitson discovered a bug in this PR regarding volume creation from another volume. I fixed that bug, and squashed that commit into previous commit. Sorry for confusion.

I can keep separate commits in PR to avoid confusion, until its ready to merge, and rebase at final stage.

clintonskitson:
@echupriyanov Very well done, LGTM. Can I get you to open the PR against the 0.3.3 branch for this one as well?

echupriyanov:
@clintonskitson I can reopen this PR against r0.3.3, but I see a problem here. PR https://github.com/emccode/rexray/pull/358 and this one can't be merged, since they are based on different AWS SDKs. I'd suggest to merge this one first, and than recreate https://github.com/emccode/rexray/pull/358 on top of it.

drivers/storage/ec2/ec2.go
```diff
@@ -921,3 +1022,17 @@ func configRegistration() *gofig.Registration {
 	r.Key(gofig.String, "", "", "", "aws.region")
 	return r
 }
+
+func awsErrCode(err error) string {
```
jdef: this func appears to be unused?

clintonskitson: good catch

clintonskitson:
@echupriyanov I agree. Let's get this over to the 0.3.3 branch and we will merge this PR so the other PR can be rebased.

drivers/storage/ec2/ec2.go
echupriyanov: Yes, its unused in this version. I was using it to parse error code in `createVolumeCreateVolume` function, but removed that call in final version.
And, while its currently unused, I think it can be useful in further development.
But if its suggested, I will remove it for clearer code.

echupriyanov:
@clintonskitson created new PR https://github.com/emccode/rexray/pull/359 against r0.3.3
I left `awsErrCode` there, but can remove it, if needed.

PR #331, opened 2016-03-02
Updated install documentation

clintonskitson:
This commit updates the documentationt to better describe how to
properly install REX-Ray manually. It also has some general
information updates.


akutz:
Damn that's good documentation. Nice.

PR #338, opened 2016-03-08
Read AWS keys from shared creds, env, or IAM

branden:
If AWS keys aren't provided in REX-Ray's config file,
`goamz.aws.GetAuth()` will fall back to a shared credentials file, then
to environment variables, and then to the ec2 instance's IAM Instance
Profile. See https://godoc.org/github.com/mitchellh/goamz/aws#GetAuth.

### DIFF
--- a/drivers/storage/ec2/ec2.go
+++ b/drivers/storage/ec2/ec2.go
@@ -47,6 +47,23 @@ func eff(fields goof.Fields) map[string]interface{} {
 	return errFields
 }
 
+func authEff(fields goof.Fields, accessKey, secretKey string) map[string]interface{} {
+	errFields := eff(map[string]interface{}{
+		"accessKey": accessKey,
+	})
+	if secretKey == "" {
+		errFields["secretKey"] = ""
+	} else {
+		errFields["secretKey"] = "******"
+	}
+	if fields != nil {
+		for k, v := range fields {
+			errFields[k] = v
+		}
+	}
+	return errFields
+}
+
 func init() {
 	core.RegisterDriver(providerName, newDriver)
 	gofig.Register(configRegistration())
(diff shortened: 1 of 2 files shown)


branden:
It looks like #325 adds this feature too. This PR is smaller and should be easier to merge, but feel free to close it in favor of the other.

clintonskitson:
Ruff! Thanks for the PR @branden. We will look at introducing either this change or the one you specified in .NEXT.

drivers/storage/ec2/ec2.go
```diff
@@ -854,9 +872,20 @@ func (d *driver) CopySnapshot(runAsync bool,
...
+		d.secretKey(),
+	)
+	auth, err := aws.GetAuth(
+		d.accessKey(),
+		d.secretKey(),
+		"",
+		time.Now(),
+	)
+	if err != nil {
+		return nil, goof.WithFields(authFields, fmt.Errorf("error getting AWS credentials: %v", err).Error())
```
akutz: Hi @branden,

Please avoid using `fmt.Errorf` and instead simply use Goof's fields and inner error function. Ex.

```
return nil, goof.WithFieldsE(authFields, "error getting AWS creds", err)
```

branden: Sure thing, thanks for the tip.

drivers/storage/ec2/ec2.go
```diff
@@ -76,10 +88,16 @@ func (d *driver) Init(r *core.RexRay) error {
 		return goof.WithFields(ef(), "error getting instance id doc")
 	}
 
-	auth := aws.Auth{
-		AccessKey: d.accessKey(),
-		SecretKey: d.secretKey(),
+	auth, err := aws.GetAuth(
```
jdef: when using IAM for credentials I think a temporary token is issued. I'm only guessing, but based on the "owning" func name here, `Init`, don't we need to periodically refresh the temporary token that AWS issues when using IAM credentials?

clintonskitson:
@branden Is there an update to this PR for timeout functionality? If so can you compare it to https://github.com/emccode/rexray/pull/325 ? We are targeting bringing in the other PR into 0.3.3 and I want to make sure that we cover what your team identified in this PR as well.

branden:
@clintonskitson The lib used by #325 refreshes credentials when they expire: https://github.com/aws/aws-sdk-go/blob/aac89266a9bc98ecbf440bb675e9dc95a94d8dc6/aws/credentials/ec2rolecreds/ec2_role_provider.go#L20

I don't intend to update this PR to add that functionality, so I'll go ahead and close it.

PR #347, opened 2016-04-01
Dockerizing REX-Ray

wlan0:
@clintonskitson I added some documentation describing the steps I followed to dockerize REX-Ray

### DIFF
--- a/.docs/user-guide/installation.md
+++ b/.docs/user-guide/installation.md
@@ -246,3 +246,47 @@ using VirtualBox's Virtual Media.
 
 A Vagrant environment and instructions using it are provided
 [here](https://github.com/emccode/vagrant/tree/master/rexray).
+
+### Containerizing REX-Ray (Dockerizing)
+Using `REX-Ray` in a docker container is a great option to quickly get it up and running. The following section discusses the steps to create and launch a docker container that launches `REX-Ray`.
+
+#### Create Docker Container
+
+Here's a Dockerfile for creating a `REX-Ray` image
+
+```
+FROM ubuntu:15.10
+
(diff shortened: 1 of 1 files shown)


Crazee:
When I tried running a containerized version of rexray with -v /var/lib/rexray:/var/lib/rexray:shared (etc.) I always get the complaint
```
docker: Error response from daemon: linux mounts: Path /var/lib/rexray is mounted on / but it is not a shared mount..
```
and haven't found a way around it yet.  This is with docker version 1.11.1.

wlan0:
You must be running docker using systemd.

If that is the case, modify /lib/systemd/system/docker.service and change the `MountFlags` mode from `slave` to `shared`

*NOTE: Sometimes, the service file can be overridden by a service file in /etc/systemd/system/docker.service*

and then execute the following steps -

```
systemctl daemon-reload
systemctl stop docker.service
systemctl start docker.service
mount --make-rshared /
```

This will allow docker to make shared mounts. Then retry the same steps you tried earlier.

Crazee:
I was trying to use a container so I wouldn't have to muck around in the underlying host, or I could have just installed rexray there.  Using docker-machine with an "amazonec2" driver magically picks the AWS AMI and installs some version of docker there.  I suppose I could manually ssh in and "fix" the two files (both /lib/systemd/system/docker.service and /etc/systemd/system/docker.service are there) but things will just break again when someone does a "docker-machine upgrade ...".  And other docker flavors (e.g. "Docker for Windows") have even less access to the VM running the underlying docker server.

Crazee:
Might try "docker-machine create --engine-install-url <magic> trump" to override the standard server config...

PR #353, opened 2016-04-06
OpenStack device discovery fix

justinclayton:
This fixes #346.

PR #357, opened 2016-04-10
Openstack device discovery fix

### LINKED ISSUE #346: Openstack driver should use /dev/disk/by-id instead of trusting Cinder
During the attach process on a Cinder volume, Nova returns hints to the OS as to what device name the volume should use. Unfortunately, most hypervisors [don't actually honor this at all](https://wiki.openstack.org/wiki/BlockDeviceConfig#Defaulting_device_names). This will cause a problem on any nova instance as soon as there is any kind of device-related hiccup, as the OS will get out of sync with where rexray believes the disk should be:

```
[root@nova1 ~]# rexray volume get --volumename rextest1
- name: rextest1
  volumeid: 3cb070d6-2e7b-4780-86e2-c9eb147753e7
  availabilityzone: nova
  status: available
  volumetype: Standard Performance
  iops: 0
  size: "40"
  networkname: ""
  attachments: []
[root@nova1 ~]# rexray volume attach --volumeid 3cb070d6-2e7b-4780-86e2-c9eb147753e7 --instanceid f16e4474-71e6-441a-a664-2c8052f7bc29
- volumeid: 3cb070d6-2e7b-4780-86e2-c9eb147753e7
  instanceid: f16e4474-71e6-441a-a664-2c8052f7bc29
  devicename: /dev/vdd
  status: ""
[root@nova1 ~]# fdisk -l
Disk /dev/vda: 34.4 GB, 34359738368 bytes, 67108864 sectors
Units = sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disk label type: dos
Disk identifier: 0x00005822
   Device Boot      Start         End      Blocks   Id  System
/dev/vda1   *        2048    67103504    33550728+  83  Linux
Disk /dev/vdb: 34.4 GB, 34359738368 bytes, 67108864 sectors
Units = sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disk /dev/vdc: 42.9 GB, 42949672960 bytes, 83886080 sectors
Units = sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
```

You can verify that `/dev/vdc` is volume id `3cd070d6...` by looking in `/dev/disk/by-id`:

```
[root@nova1 by-id]# ls -ld /dev/disk/by-id/virtio-3cb070d6-2e7b-4780-8
lrwxrwxrwx. 1 root root 9 Mar 29 19:05 /dev/disk/by-id/virtio-3cb070d6-2e7b-4780-8 -> ../../vdc
```

According to my Slack conversation with @clintonskitson, `/dev/disk/by-id` is used by other drivers, so the recommendation here would probably be to use the same method (especially since it's the only foolproof method we have with Openstack at the moment).

Thanks!

justinclayton:
Fixes #346. Reopen of #353, but now merging against [release/0.3.3](https://github.com/emccode/rexray/tree/release/0.3.3) branch.

### DIFF
--- a/drivers/storage/openstack/openstack.go
+++ b/drivers/storage/openstack/openstack.go
@@ -319,6 +319,8 @@ func (d *driver) getVolume(
 		"volumeId":   volumeID,
 		"volumeName": volumeName})
 
+	log.WithFields(fields).Info("d.getVolume")
+
 	if volumeID != "" {
 		volume, err := volumes.Get(d.clientBlockStorage, volumeID).Extract()
 		if err != nil {
@@ -382,7 +384,8 @@ func (d *driver) GetVolume(
 			attachmentSD := &core.VolumeAttachment{
 				VolumeID:   attachment["volume_id"].(string),
 				InstanceID: attachment["server_id"].(string),
-				DeviceName: attachment["device"].(string),
+				// never trust the cinder-assigned device name; use /dev/disk/by-id
+				DeviceName: getAttachedDevicePathByID(attachment["volume_id"].(string)),
 				Status:     "",
 			}
 			attachmentsSD = append(attachmentsSD, attachmentSD)
@@ -413,6 +416,8 @@ func (d *driver) GetVolumeAttach(
 		"instanceId": instanceID,
 	})
 
+	log.WithFields(fields).Info("d.GetVolumeAttach")
+
 	if volumeID == "" {
 		return []*core.VolumeAttachment{},
 			goof.WithFields(fields, "volumeId is required")
@@ -427,6 +432,8 @@ func (d *driver) GetVolumeAttach(
 		var attached bool
 		for _, volumeAttachment := range volume[0].Attachments {
 			if volumeAttachment.InstanceID == instanceID {
+				// never trust the cinder-assigned device name; use /dev/disk/by-id
+				volumeAttachment.DeviceName = getAttachedDevicePathByID(volumeAttachment.VolumeID)
 				return volume[0].Attachments, nil
 			}
 		}
@@ -763,42 +770,11 @@ func (d *driver) RemoveVolume(volumeID string) error {
 }
 
 func (d *driver) GetDeviceNextAvailable() (string, error) {
-	letters := []string{"a", "b", "c", "d", "e", "f", "g", "h", "i", "j", "k", "l", "m", "n", "o", "p"}
-	blockDeviceNames := make(map[string]bool)
+	return "", errors.ErrNotImplemented
+}
 
-	blockDeviceMapping, err := d.GetVolumeMapping()
-	if err != nil {
-		return "", err
-	}
-
-	for _, blockDevice := range blockDeviceMapping {
-		re, _ := regexp.Compile(`^/dev/vd([a-z])`)
-		res := re.FindStringSubmatch(blockDevice.DeviceName)
-		if len(res) > 0 {
-			blockDeviceNames[res[1]] = true
-		}
-	}
-
-	localDevices, err := getLocalDevices()
-	if err != nil {
-		return "", err
-	}
-
-	for _, localDevice := range localDevices {
-		re, _ := regexp.Compile(`^vd([a-z])`)
-		res := re.FindStringSubmatch(localDevice)
-		if len(res) > 0 {
-			blockDeviceNames[res[1]] = true
-		}
-	}
-
-	for _, letter := range letters {
-		if !blockDeviceNames[letter] {
-			nextDeviceName := "/dev/vd" + letter
-			return nextDeviceName, nil
-		}
-	}
-	return "", goof.New("No available device")
+func getAttachedDevicePathByID(volumeID string) string {
+	return fmt.Sprintf("/dev/disk/by-id/virtio-%s", volumeID[:20])
 }
 
 func getLocalDevices() (deviceNames []string, err error) {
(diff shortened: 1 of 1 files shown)


clintonskitson:
@justinclayton Was spending some quality time with this one and noticed a problem around the use or lack of `/dev/disk/by-id` in centos systems in OpenStack. It looks as though the necessary information isn't coming down through the kernel drivers. Can you take a peak at this and see if you run into the same problems?

justinclayton:
@clintonskitson I tested exclusively on instances running Centos 7.2.1511.

```
# uname -a
Linux m1-slave-16 3.10.0-327.el7.x86_64 #1 SMP Thu Nov 19 22:10:57 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

Can you elaborate on what you're seeing?

clintonskitson:
Can you try other revs of centos?

I am seeing that the /dev/disk/by-id is not showing up.

clintonskitson:
Btw. I am testing on Rackspace. If you have an account you should be able
to fire up a CentOS instance easily and see the directory is missing.

Been trying to figure out the package or update that would bring that
directory in from udev but have been unable to nail that down.

justinclayton:
On CentOS 6 and 7, It looks like the `/dev/disk/by-id` is driven by udev. Here's the snippet from `/lib/udev/rules.d/60-persistent-storage.rules`:

```
# virtio-blk
KERNEL=="vd*[!0-9]", ATTRS{serial}=="?*", ENV{ID_SERIAL}="$attr{serial}", SYMLINK+="disk/by-id/virtio-$env{ID_SERIAL}"
KERNEL=="vd*[0-9]", ATTRS{serial}=="?*", ENV{ID_SERIAL}="$attr{serial}", SYMLINK+="disk/by-id/virtio-$env{ID_SERIAL}-part%n"
```

I can't speak for Rackspace, but in my openstack environment, the first 20 characters of the Cinder volume id are added to the serial of the attached device:

```# cat /sys/block/vdd/serial
f2d1c222-993a-4e94-b```

Which is what triggers udev to create this symlink:

```
# ls -l /dev/disk/by-id/virtio-*
lrwxrwxrwx. 1 root root 9 Apr 15 18:37 /dev/disk/by-id/virtio-f2d1c222-993a-4e94-b -> ../../vdd
```

Don't know if it's useful, but for completeness, here's the dmesg output for the attachment as well:

```
65076.733195] ACPI: \_SB_.PCI0.S48_: ACPI_NOTIFY_DEVICE_CHECK event
[65076.733332] ACPI: \_SB_.PCI0.S48_: Device check in hotplug_event()
[65076.733653] pci 0000:00:09.0: [1af4:1001] type 00 class 0x010000
[65076.734213] pci 0000:00:09.0: reg 0x10: [io  0x0000-0x003f]
[65076.734303] pci 0000:00:09.0: reg 0x14: [mem 0x00000000-0x00000fff]
[65076.735848] pci 0000:00:09.0: BAR 1: assigned [mem 0xc0000000-0xc0000fff]
[65076.739678] pci 0000:00:09.0: BAR 0: assigned [io  0x1000-0x103f]
[65076.743990] virtio-pci 0000:00:09.0: enabling device (0000 -> 0003)
[65076.754415] virtio-pci 0000:00:09.0: virtio_pci: leaving for legacy driver
[65076.761238] virtio-pci 0000:00:09.0: irq 33 for MSI/MSI-X
[65076.761279] virtio-pci 0000:00:09.0: irq 34 for MSI/MSI-X
[65076.782783]  vdd: unknown partition table
```

I hope that helps us figure out what's different about the way Rackspace is doing this. My guess is that if they don't populate the serial, that could account for why /dev/disk/by-id/virtio-x isn't showing up.

Let me know how else I can help.

clintonskitson:
Yeah I am still stuck on getting this equivalent working in Rackspace. Which hypervisor are you using?

justinclayton:
KVM

justinclayton:
What does /sys/block/vdX/serial say on your side?

clintonskitson:
`serial` is not present.

clintonskitson:
I believe it is Xen that this instance is running on.

justinclayton:
So, Xen isn't actually affected by #346 anyway. Maybe this should only be a fallback behavior for non-Xen nodes. Thoughts? I can put that together in a different PR if we want to go that route.

clintonskitson:
Interesting, so was that based on the research you did in the openstack docs that Xen wasn't impacted?

justinclayton:
Yes, we've had to deal with this for a while so we were already familiar: https://bugs.launchpad.net/nova/+bug/1075971

Xen is able to reliably take the device names assigned by Cinder during attach, while hypervisors that use libvirt can only take those device ordering hints at boot time, which obviously doesn't work for live-attaching volumes. If you were to make two attach calls in rapid succession, the only way I'm aware of to tell with certainty which device is which is to match the serial to the Cinder volume id.

justinclayton:
Actually, upon reading through some other related bugs I've realized that the virtio stuff is only going to work for libvirt anyway, so we're almost certainly going to have to have both behaviors in play. :-(

clintonskitson:
So a detect hypervisor step with dmidecode and then a choice between doing
methodbofsl specifying to xen the next device or doing the dynamic
discovery for kvm?

justinclayton:
Yeah that sounds good. Because I need to put the old code back in, I may delete and recreate this branch.

PR #359, opened 2016-04-12
Replace goamz with aws-sdk-go (rebased against release/0.3.3)

echupriyanov:
This is a direct rewrite using aws-sdk-go.
Some minor tweaks:
* rexray now picks AWS credentials from all known places (environment,
  ~/.aws/credentials and EC2 instance profile)
* volume attach - API call _requires_ presence of instanceID.
  If missing, RexRay nevertheless passes null instance ID to AWS getting an error.
  Now driver's EC2 instance ID is used, if none provided.
* snapshot copy - as with volume attach, destinationregion is mandatory. And if its omitted,
  AWS returns an error. And as with volume copy, driver's EC2 instance region is used as default

Everything else should work as before

Functional tests are passed

Fixed volume create from another volume

Added AWS IAM policy to documentation

Signed-off-by: Eugene Chupriyanov <tchu@tchu.ru>

### DIFF
--- a/.docs/user-guide/storage-providers.md
+++ b/.docs/user-guide/storage-providers.md
@@ -11,21 +11,28 @@ This page reviews the storage providers and platforms supported by `REX-Ray`.
 The Amazon EC2 driver registers a storage driver named `ec2` with the `REX-Ray`
 driver manager and is used to connect and manage storage on EC2 instances. The
 EC2 driver is made possible by the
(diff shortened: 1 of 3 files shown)


clintonskitson:
@akutz This one is good to go.

kacole2:
Fantastic work! :+1+

PR #366, opened 2016-04-13
Added a disable feature for create/remove volume

clintonskitson:
This commit introduces a new option to make sure the create/remove
volume functionality can be disabled from the volume and
integration plugins.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -682,6 +682,30 @@ The above example disables the `default-docker` and `isilon3` modules such that
 This section describes various global configuration options related to
 operations such as mounting and unmounting volumes.
 
+#### Disable Create
+The disable create feature enables you to disallow any volume creation activity.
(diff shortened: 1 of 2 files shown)


miry:
@clintonskitson is it released ? and how can we use it?

akutz:
Hi @miry,

It means it's merged into the release train that we're trying to push to RC later today. As for when, as an RC perhaps in the next 18 hours. As a release, in the next week.

miry:
@akutz thanks. about remove, it means rexray would not remove named volumes, but will remove unnamed?

clintonskitson:
@miry This means RR at no time removes any volumes if configured in this way. Today there is no differentiation in the API from Docker between a named or anonymous volume.

clintonskitson:
This is configurable through the module or globally so it is intended to block Docker or Mesos users from these operations.

PR #370, opened 2016-04-14
Fixed insecure option for openstack

clintonskitson:
The insecure option was added in this commit which was never
    properly imeplemented. Additionally updated availzone param
    in docs to align to common configs.


docwhat:
Was this actually ever merged? Because I don't see it in the git commit history for `v0.3.3` or `master`.

Issue #365, opened 2016-04-13
Debug logging not occurring with Ubuntu 14.04


clintonskitson:
@akutz It looks like with RR 0.3.2 and Ubuntu 14.04, the logging is not making it to `/var/log/rexray/rexray.log`. Running it in the foreground does yield the necessary information. Can you review with an Ubuntu instance?

`REXRAY_LOGLEVEL=debug`

```

                          ⌐▄Q▓▄Ç▓▄,▄_
                         Σ▄▓▓▓▓▓▓▓▓▓▓▄π
                       ╒▓▓▌▓▓▓▓▓▓▓▓▓▓▀▓▄▄.
                    ,_▄▀▓▓ ▓▓ ▓▓▓▓▓▓▓▓▓▓▓█
                   │▄▓▓ _▓▓▓▓▓▓▓▓▓┌▓▓▓▓▓█
                  _J┤▓▓▓▓▓▓▓▓▓▓▓▓▓├█▓█▓▀Γ
            ,▄▓▓▓▓▓▓^██▓▓▓▓▓▓▓▓▓▓▓▓▄▀▄▄▓▓Ω▄
            F▌▓▌█ⁿⁿⁿ  ⁿ└▀ⁿ██▓▀▀▀▀▀▀▀▀▀▀▌▓▓▓▌
             'ⁿ_  ,▄▄▄▄▄▄▄▄▄█_▄▄▄▄▄▄▄▄▄ⁿ▀~██
               Γ  ├▓▓▓▓▓█▀ⁿ█▌▓Ω]█▓▓▓▓▓▓ ├▓
               │  ├▓▓▓▓▓▌≡,__▄▓▓▓█▓▓▓▓▓ ╞█~   Y,┐
               ╞  ├▓▓▓▓▓▄▄__^^▓▓▓▌▓▓▓▓▓  ▓   /▓▓▓
                  ├▓▓▓▓▓▓▓▄▄═▄▓▓▓▓▓▓▓▓▓  π ⌐▄▓▓█║n
                _ ├▓▓▓▓▓▓▓▓▓~▓▓▓▓▓▓▓▓▓▓  ▄4▄▓▓▓██
                µ ├▓▓▓▓█▀█▓▓_▓▓███▓▓▓▓▓  ▓▓▓▓▓Ω4
                µ ├▓▀▀L   └ⁿ  ▀   ▀ ▓▓█w ▓▓▓▀ìⁿ
[...]

PR #373, opened 2016-04-15
Fixed logging not occuring with Ubuntu 14.04

philippfranke:
This patch is a fix for issue #365, where the logging is not writing
to `/var/log/rexray/rexray.log`.

Since os.Stderr is attached to TTY(`!log.IsTerminal()`) rexray doesn't
write to the log file so I removed `cmd.Stderr`.

### DIFF
--- a/rexray/cli/service.go
+++ b/rexray/cli/service.go
@@ -231,7 +231,6 @@ func (c *CLI) tryToStartDaemon() {
 	}
 
 	cmd := exec.Command(thisAbsPath, cmdArgs...)
-	cmd.Stderr = os.Stderr
 
 	cmdErr := cmd.Start()
 	failOnError(cmdErr)


akutz:
Hi @philippfranke,

I'm not sure we want to do this. I based this detection on the same detection in logrus, which recently [updated its logic](https://github.com/Sirupsen/logrus/pull/303/files). Additionally, this apparently only happens on Ubuntu 14.04, so unless this patch has been tested this on all other, supported distros, this patch won't be accepted that will have possible side-effects for other distributions.

clintonskitson:
Definitely need to look into this one further. Thanks for the feedback so far and work towards a fix @philippfranke.

clintonskitson:
@akutz I tested this and I believe the fix does in fact take care of the issue and doesn't seem to represent any detrimental issues. Your call.

akutz:
Hi @philippfranke / @clintonskitson,

I can cherry-pick @philippfranke's fix or he can close and reopen this PR against the `release/0.3.3` branch. I could manually pull the patch over, but a merge would overwrite other bits on the `release/0.3.3` branch as well. @philippfranke, it's your call here -- either reopen this PR against 0.3.3 or I can cherry-pick your commit across, and we'll just reflect in this PR what occurred.

PR #375, opened 2016-04-16
Updated goscaleio for re-auth loop fix

clintonskitson:
This commit adds a fix for ScaleIO where bad creds would cause
a re-authentication loop.


akutz:
Hi @clintonskitson,

If this is a goscaleio patch, why are there changes in the ec2 driver's imports section?

akutz:
Eh, it appears to be a formatting issue.

PR #377, opened 2016-04-17
Fixed logging not occuring with Ubuntu 14.04

philippfranke:
This patch is a fix for issue #365, where the logging is not writing
to /var/log/rexray/rexray.log.

Since os.Stderr is attached to TTY(!log.IsTerminal()) rexray doesn't
write to the log file so I removed cmd.Stderr.

Follow up: https://github.com/emccode/rexray/pull/373


akutz:
Thank you @philippfranke, we very much appreciate it! (cc @clintonskitson)

PR #382, opened 2016-04-20
WiP - libStorage Types

akutz:
This patch replaces REX-Ray's core model with the libStorage equivalent types.

### Framework
Update the framework with the types.
- [x] Core
- [ ] Tests

### Drivers
Update the drivers with the types.
- [x] mock
- [x] ec2
- [x] gce
- [ ] isilon
- [ ] openstack
- [ ] rackspace
- [ ] scaleio
- [ ] virtualbox
- [ ] vmax
- [ ] xtremio

@clintonskitson - One caveat, we need to make changes to each driver so that the `StartTime` for snapshots is a timestamp format we know how to parse as it's stored as a `int64` in libStorage's `Snapshot` object, permitting the time to be stored as an locale-agnostic UNIX epoch value. However, to arrive at that value we'll need to understand the values being returned by each driver so we can parse them into a golang time structure and then emit the UNIX epoch.


akutz:
@clintonskitson, can you please provide feedback on the question raised in the PR description? Thanks!

clintonskitson:
Interesting question. It does seem to be tied to logic that must take place
in each driver. My assumption is that the difficulty level shouldn't be
high as each driver should be representing time with timezone associated if
they do have time.

akutz:
Hi @clintonskitson,

It's not just a timezone issue, but a formatting issue since if a storage platform returns time information as a string, we need to know the format of that string so we can parse it into a `time.Time` struct and then emit `time.Unix()` to get the epoch.

PR #403, opened 2016-05-02
Feature/docker volume mountpath

clintonskitson:
This commit introduces a new configurable field for
docker.volume.mountPath to choose the default location of
mounted volumes. A path of /var/lib/rexray/volumes is still
default as a base.

This has a commit from @akutz below this one.

### DIFF
--- a/rexray/cli/cli.go
+++ b/rexray/cli/cli.go
@@ -319,6 +319,10 @@ func (c *CLI) preRun(cmd *cobra.Command, args []string) {
 
 	c.updateLogLevel()
 
+	if !c.config.IsSet("docker.mountDirPath") {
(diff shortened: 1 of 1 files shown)


rexray/cli/cli.go
```diff
@@ -347,6 +347,10 @@ func (c *CLI) preRun(cmd *cobra.Command, args []string) {
 		if c.runAsync {
 			c.ctx = c.ctx.WithValue("async", true)
 		}
+		if !c.config.IsSet("docker.mountDirPath") {
+			c.config.Set("docker.mountDirPath", "/var/lib/rexray/volumes")
```
akutz: Hi @clintonskitson,

The value should be `util.LibFilePath("volumes")` where `util` is the package `github.com/emccode/rexray/util`

PR #407, opened 2016-05-05
Fixed cli commands to for nil exceptions

clintonskitson:
The CLI commands were fixed to ensure the proper store option
is being passed.

### DIFF
--- a/rexray/cli/cmds_volume.go
+++ b/rexray/cli/cmds_volume.go
@@ -167,7 +167,10 @@ func (c *CLI) initVolumeCmds() {
 
 			vol, _, err := c.r.Storage().VolumeAttach(
 				c.ctx, c.volumeID,
-				&apitypes.VolumeAttachOpts{Force: c.force})
+				&apitypes.VolumeAttachOpts{
+					Force: c.force,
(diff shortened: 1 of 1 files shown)


akutz:
Hi @clintonskitson,

I appreciate you correcting this, but OOC, why replace `store()` with `utils.NewStore()` when the former is just an alias of the latter, defined in [cli.go](https://github.com/clintonskitson/rexray/blob/bugfix/cli_store_opts/rexray/cli/cli.go#L472).

clintonskitson:
@akutz Most of those were net new to fix nil exceptions. I will translate it to `store()`.

akutz:
Hi @clintonskitson,

I think some were new, but others you either changed or added between my commit to fix the same thing. See the PR's diff where it shows that `store()` was already in place in several locations:

image

clintonskitson:
Agreed, fixing PR to go back to `store()` and add it where needed to avoid nils.

akutz:
Hi @clintonskitson,

The only reason I care at all is that `store()` is used in the other files that require the same value.

clintonskitson:
@akutz This one should be good.

PR #409, opened 2016-05-05
Updated the install script

clintonskitson:
This commit updates the install script to fix a problem relating
to the URL generated when specifying a specific version to
install.


clintonskitson:
This won't work until builds are generating versioned deb/rpm files.

PR #420, opened 2016-05-09
Fix for starting libs client from module

clintonskitson:
This commit makes it so that libstorage client is started once
per module. This allows you to configure libs client settings
including volume and integration settings per module.

### DIFF
--- a/daemon/module/module.go
+++ b/daemon/module/module.go
@@ -167,7 +167,7 @@ func InitializeDefaultModules() error {
 			util.LibFilePath("volumes"))
 	}
 
-	lsc, _, err, errs := libstorage.New(c)
+	_, _, err, errs := libstorage.New(c)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


akutz:
Hi @clintonskitson,

This could be an issue. The line `lsc, _, err, errs := libstorage.New(mc.Config)` actually attempts to spin up a server if no host is detected. I think we should leave lines [164](https://github.com/emccode/rexray/pull/420/files#diff-8a14c14c89038f809f618c5a5b701d23L164) - [174](https://github.com/emccode/rexray/pull/420/files#diff-8a14c14c89038f809f618c5a5b701d23L174) and then update the new code to use the `Dial` method instead of the `New` method. The `Dial` method *only* attempts to connect to an existing host.

clintonskitson:
@akutz Will take a peak here.

PR #429, opened 2016-05-14
Feature/module startup timeout

clintonskitson:
This includes multiple commits. The first are minor tweaks that I believe are relevant to changes in the context with libs and metadata. The other is a new configurable setting for module timeout so that longer running pathcache/volume list+attachments operations do not cause the modules startup to fail.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,7 +1,7 @@
 package: github.com/emccode/rexray
 import:
   - package: github.com/emccode/libstorage
-    ref:     v0.1.0-rc2
+    ref:     8d0aaaec8cc523697e6ee8de80a22585a4ab58e1
   - package: github.com/Sirupsen/logrus
     ref:     feature/logrus-aware-types
     repo:    https://github.com/akutz/logrus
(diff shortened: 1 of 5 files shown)


glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/emccode/rexray
 import:
   - package: github.com/emccode/libstorage
-    ref:     v0.1.0-rc2
+    ref:     8d0aaaec8cc523697e6ee8de80a22585a4ab58e1
```
akutz: Hi @clintonskitson,

Please omit the ref altogether for now, at least until we tag LS RC 0.1.0-RC3. That way this build can pull in a new LS commit if we patch it.

clintonskitson: Rgr

clintonskitson:
@akutz Updated.

glide.yaml
akutz: Hi @clintonskitson,

I meant please omit the ref completely. It will default to the master branch if nothing is specified, and that's what we want for now.

PR #444, opened 2016-05-26
change uname -p to uname -m in Makefile

cduchesne:
changing uname -p to uname -m. uname -p is not portable between operating systems and -m provides the expected result


akutz:
Hi @cduchesne,

Can you please rebase this PR onto the current tip of `master`. PRs should have one commit in most cases. Thanks!

cduchesne:
@akutz - There was only 1 commit when I submitted. It should be fixed now.

PR #456, opened 2016-06-09
Allow RexRay Http Client Timeout Configurable

Oskoss:
When running libstorage with our scaleIO setup sometimes mount/unmount takes longer than 10 seconds. With this PR we introduce an extra config parameter that can be set in the configuration.yml file to adjust the timeout. The default will remain at 10 seconds.

### DIFF
--- a/daemon/module/docker/volumedriver/voldriver.go
+++ b/daemon/module/docker/volumedriver/voldriver.go
@@ -152,11 +152,13 @@ func (m *mod) Start() error {
 	} else {
 		specPath = addr
 		startFunc = func() error {
+			readTimeout := m.config.GetInt("http.readtimeout")
+			writeTimeout := m.config.GetInt("http.writetimeout")
 			s := &http.Server{
 				Addr:           addr,
 				Handler:        mux,
-				ReadTimeout:    10 * time.Second,
-				WriteTimeout:   10 * time.Second,
+				ReadTimeout:    time.Duration(readTimeout) * time.Second,
+				WriteTimeout:   time.Duration(writeTimeout) * time.Second,
 				MaxHeaderBytes: 1 << 20,
 			}
 			return s.ListenAndServe()
--- a/daemon/module/module.go
+++ b/daemon/module/module.go
@@ -105,6 +105,9 @@ rexray:
             desc:     The default docker module.
             host:     unix:///run/docker/plugins/rexray.sock
             spec:     /etc/docker/plugins/rexray.spec
+            http:
+              writetimeout: 10
+              readtimeout: 10
             disabled: false
 `)
 	cfg.Key(gofig.String, "", "10s", "", "rexray.module.startTimeout")


akutz:
Hi Peter,

Really? REALLY? You of all people should know not here :) This is a good patch, and we should add it, but @Oskoss, you know where it should be added :)

akutz:
Hi @Oskoss,

You'll want to patch in the config options [here](https://github.com/emccode/libstorage/blob/master/drivers/storage/libstorage/libstorage_driver.go#L67).

akutz:
Also, this isn't even necessarily a thing in the new version since the server-side op occurs on a separate goroutine from the HTTP request.

clintonskitson:
This is a really interesting find. @akutz How can we properly prepare for this? This is causing a situation where the mount count is increased on subsequent retry requests.

1) is our count increment happening in the right place?
2) how does our internal timeout on libs relate?
3) how would a timeout that we force within the mount/unmount operation relate to what gets set here? https://github.com/emccode/libstorage/issues/180

akutz:
Hi @clintonskitson,

It's actually not new. If you recall we ran into it last year with the exact same driver, ScaleIO, and I increased the timeout on the Dublin branch to 300 seconds (5 min) for this very reason.

Remember all the times you asked me why the task service? Well, this is the answer I've given you every time. 

Nothing to see here, move on...essentially. It's an issue of which I'm already aware and for which I've already prepared.

akutz:
I'm closing this thread. If anyone wants to continue the discussion, please re-open it on the libStorage tracker. @clintonskitson - please remember that what the server does and how the client reacts are two different things and are considered separately.

clintonskitson:
@akutz This PR seems to be RR specific as RR is what advertises the UNIX socket to the consuming service (Docker). What happens in libStorage is important to align. Is there somewhere else to turn this thread to for better awareness of planned fix? Seems like a configurable option is a valid thing to consider for now.

1) RR module timeout
2) libStorage HTTP timeout
3) libStorage registry level operation timeout (mount/unmount)
3) libStorage storage driver operation timeout

akutz:
@clintonskitson / @Oskoss - You're both correct; I apologize. I misread the diff and thought it was a different HTTP server. We have more than one. I blame the vodka I started drinking with lunch. *sigh*.

akutz:
@clintonskitson - So what do you want to do? I'll merge this, but looking forward there are a few possibilities:

- Move the Docker Volume Endpoint (DVE) into libStorage like I mentioned awhile back
- Allow drivers to all advertise their recommended timeouts and use the max value from all of those to configure the timeout duration for the DVE in REX-Ray
- Remove any timeout whatsoever

Honestly, I'm kind of feeling the last one. I think we should consider the DVE a dumb proxy at this point and defer all logic to upstream components. Thoughts?

Oskoss:
My opinion which is worth 1 cent :)

- Keep DVE separate from libStorage. I like that architecture and the structure no need to dilute it with the DVE. 
- Having drivers advertise the timeouts is a pain to go back and refactor....possible but painful.
- Removing timeouts is my go to, the default for golang is to just have infinite timeouts so why wouldn't we? 

Thanks for the last comment @akutz made my day haha

akutz:
Hi @Oskoss,

Your opinion is quite valued, even if you downplay it. I agree with your assessment, and I'll likely remove the timeouts and close this issue. @clintonskitson, this will defer all the above questions to libStorage, which I believe is desirable.

clintonskitson:
@akutz I agree with the removal of the timeouts on this endpoint. Derivative calls and logic should handle success/failures/timeouts.

akutz:
Hi @clintonskitson,

Okay, so I'm going to remove the timeout, but something occurred to me -- this is the server timeout. I did some digging, and it looks like Docker's [client timeout is 30 seconds](https://github.com/docker/docker/blob/master/pkg/plugins/client.go). I haven't traced it, but do you know if they expect things to return immediately for the Volume plug-in, or if they simply submit an op and then check every now and then for the result?

clintonskitson:
@akutz on Docker's side I believe it is a synchronous event where they expect the return from the operation immediately. There is then a backoff delay for retries.

akutz:
That's what I saw in that same source file. I just traced it down, and check it out. The [transport gets created](https://github.com/docker/docker/blob/master/pkg/plugins/client.go#L24-L41) and then it is configured to [timeout at 32 seconds](https://github.com/docker/go-connections/blob/master/sockets/sockets.go).

akutz:
So if it times out and continues to call and retry over and over, I haven't looked at the Docker Volume Driver on the REX-Ray side, but it seems your concerns about whether we will continue to retry the same operation over and over could be valid if we don't detect the retry attempt from Docker. We're not currently scoped time-wise to handle this issue as part of this release as it's never something I've looked at up to this point. Even last year with Dublin we didn't run into this issue as the one we did encounter was with what eventually became the libStorage server.

clintonskitson:
@akutz Ok considering that, us not specifying a timeout here probably makes sense. Considering the thread here, a future end solution here is to start tracking the ID that comes through the Mount() operation.

```
{
    "Name": "volume_name",
    "ID": "b87d7442095999a92b65b3d9691e697b61713829cc0ffd1bb72e4ccd51aa4d6c"
}
```

With this could track unique mount requests I believe, which would make the operation idempotent for us based on the same initiating request from Docker.

clintonskitson:
Docker plugin docs https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md

akutz:
Or we simply ignore this entirely and it will go away once we deprecate this integration pattern with Docker. Sure, let's increase the timeout, but as for tracking the ID, IMO we should focus on Docker's new, favored integration solution.

PR #478, opened 2016-06-28
Fix squashing the error (vol = nil) when doing a volume create

dvonthenen:
The issue https://github.com/emccode/polly/issues/88 was reported in Polly, but the root cause exists in rexray.

```
root@node1:/tmp# rexray volume create --volumename=haha
FATA[0001] error creating new volume                     inner.size=0 inner.volumeName=haha inner.provider=virtualbox status=500
```

### DIFF
--- a/rexray/cli/cmds_volume.go
+++ b/rexray/cli/cmds_volume.go
@@ -96,10 +96,6 @@ func (c *CLI) initVolumeCmds() {
 		Aliases: []string{"new"},
 		Run: func(cmd *cobra.Command, args []string) {
 
-			if c.size == 0 && c.snapshotID == "" && c.volumeID == "" {
-				log.Fatalf("missing --size")
-			}
-
 			opts := &apitypes.VolumeCreateOpts{
 				AvailabilityZone: &c.availabilityZone,
 				Size:             &c.size,
@@ -119,7 +115,7 @@ func (c *CLI) initVolumeCmds() {
 			} else if c.snapshotID != "" && c.volumeName != "" {
 				volume, err = c.r.Storage().VolumeCreateFromSnapshot(
 					c.ctx, c.snapshotID, c.volumeName, opts)
-			} else if c.volumeName != "" {
+			} else {
 				volume, err = c.r.Storage().VolumeCreate(
 					c.ctx, c.volumeName, opts)
 			}


clintonskitson:
@dvonthenen I am not quite sure I understand the fix here. Can you provide the result of the fix that you applied in terms of the response after the fix?

It just doesn't make sense that this would provide the fix for the squashing errors problem.

akutz:
Hi @clintonskitson,

It's not about squashing errors actually. The title is misleading. Let's jump on a hangout?

dvonthenen:
In a nutshell, the errors sent back are expected to be an object serialized in JSON. The issue is we never get a volume object which in turn only gives us a generic error in non-object form (just a string error). The deserialization fails on the client which gives us the "http error".

If that still doesnt make any sense, we can talk about it on hangouts later like Andrew was trying to organize.

akutz:
Since @clintonskitson isn't available to V-chat at the moment, I'll do my best to explain the issue here. Clint, what David and I found was that the original logic of the Create command was preventing the remote Create call if the volume name was not provided:

```
} else if c.volumeName != "" {
```

The issue is that if the remote call is not performed then neither the `err` or `volume` variables are assigned.

Thus since there is no error, [this](https://github.com/dvonthenen/rexray/blob/e7176b3a3422ffb9d42a97f79991140aa18b97d9/rexray/cli/cmds_volume.go#L123-L131) is hit:

```
			if err != nil {
				log.Fatal(err)
			}

			out, err := c.marshalOutput(&volume)
			if err != nil {
				log.Fatal(err)
			}
			fmt.Println(out)
```

The address of the `volume` variable is marshalled to YAML/JSON and emitted to Stdout. Since `nil` is a perfectly valid value to marshal, and it's marshaled as the string `<nil>`, that's what was being printed and displayed to the user. Here's [an example](https://play.golang.org/p/x3KVDg74JR) on the Golang Playground:

```
package main

import (
	"fmt"
)

func main() {
	fmt.Println(nil)
}
```

The above code prints `<nil>` to Stdout.

What @dvonthenen and I decided was that we should only do variable and flag checking locally *IFF* it is used to determine the remote route. For example:

```
			if c.volumeID != "" && c.volumeName != "" {
				volume, err = c.r.Storage().VolumeCopy(
					c.ctx, c.volumeID, c.volumeName, opts.Opts)
			} else if c.snapshotID != "" && c.volumeName != "" {
				volume, err = c.r.Storage().VolumeCreateFromSnapshot(
					c.ctx, c.snapshotID, c.volumeName, opts)
			} else {
				volume, err = c.r.Storage().VolumeCreate(
					c.ctx, c.volumeName, opts)
			}
```

In the above snippet if the `volumeID` and `volumeName` are both set the `VolumeCopy` route is used. If the `snapshotID` and `volumeName` fields are set then the `VolumeCreateFromSnapshot` route is used instead.

The final condition is simply an `else` block that defaults to the `VolumeCreate` route. If the `volumeName` value is not set then the **_REMOTE_** server will send back the appropriate error, and it will be displayed to the user. 

tl;dr - the local client should not be responsible for input validation unless it affects the route; that job belongs to the libStorage API.

akutz:
@dvonthenen - Your [comment](https://github.com/emccode/rexray/pull/478#issuecomment-229411802) isn't quite accurate. The issue isn't anything to do with not getting a volume object -- the issue is what I described above -- the logic to create a volume was simply never being invoked since we were validating a the `volumeName` locally instead of deferring that to the server.

Anyway, this code should fix the issue.

Issue #471, opened 2016-06-22
Rexray 0.3.3 config file not compatible with Rexray 0.4 


VibhorGupta:
Rexray 0.3.3 config file is not compatible with Rexray 0.4. 

I have 2 instances of rexray 0.3.3 running fine. I pulled rexray 0.4 on a new server and used the same config file as on the 0.3.3.

rexray 0.3.3 config file example:

```
rexray:
storageDrivers:
- scaleio
volume:
 mount:
  preempt: true
scaleio:
endpoint: xxxxxxxxx
insecure: true
userName: admin
password: xxxxxxxxx
systemID: 0
protectionDomainName: pd1
storagePoolName: hdd
```

Rexray 0.4 was dying with the following error:

```

rexray service start -l debug
DEBU[0000] updated log level                             logLevel=debug
DEBU[0000] os.args                                       time=1466613572 val=[rexray service start -l debug]
DEBU[0000] invoking service start                        os.Args=[rexray service start -l debug] time=1466613572
DEBU[0000] starting in background                        time=1466613572
Starting REX-Ray...DEBU[0000] trying to start service                       client=/var/run/rexray/client.sock time=1466613572
DEBU[0000] accepted connection                           time=1466613573
DEBU[0000] received data                                 time=1466613573
FAILED!
DEBU[0000] exiting with error code 1
```

temporary solution is to use the following config file:

```
libstorage:
  service: scaleio
  integration:
    volume:
      operations:
        mount:
          preempt: true
scaleio:
  endpoint: xxxxxxxxxx
  insecure: true
  userName: admin
  password: xxxxxxxxxx
  systemID: 0
  protectionDomainName: pd1
  storagePoolName: hdd
```

it would be good to use the old rexray.storagedrivers as default if libstorage.service is not defined.

thanks

akutz:
Thank you for taking the time to file this @vibhorgupta! I'm glad we got you online again so quickly. I'll definitely mark this patch a priority and ensure it's included in 0.4.1.

akutz:
Hi @VibhorGupta, I tested #481 and showed it solves this, so I'm going to close this issue. Look for this fix in the next release.

PR #481, opened 2016-06-29
REX-Ray Config StorageDrivers Backwards Compat

akutz:
This patch should provide backwards compatibility with the old property `rexray.storageDrivers`. If this property is set, its first element will be assigned to the property `libstorage.service` if that property is undefined. This is regarding issue #471.

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -326,6 +326,19 @@ func ActivateLibStorage(
 
 	config = config.Scope("rexray")
 
+	// set the `libstorage.service` property to the value of
+	// `rexray.storageDrivers` if the former is not defined and the
+	// latter is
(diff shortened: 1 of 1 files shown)


akutz:
@clintonskitson Can you think of any other properties we should include? This seemed to most obvious. I'm not going to bother with the `volumeDrivers` or `osDrivers` properties since those were never set to anything other than the defaults in previous versions of RR.

clintonskitson:
@akutz I believe we have backward compat built in for other options. As long we are verbosely describing the general options that were set in the log to cover that scenario then I believe the backward compatibility based primarily on storage driver makes sense.

PR #491, opened 2016-07-06
Added REX-Ray Configuration Generator to docs

kacole2:
Added the REX-Ray Configuration Generator to the Docs so it will be a quick way to get going without diving too deep into reading. In addition, added a QuickConfig section to the Configuration page.

Signed-off-by: Kendrick Coleman kendrickcoleman@gmail.com

### DIFF
--- a/.docs/index.md
+++ b/.docs/index.md
@@ -104,24 +104,21 @@ on building REX-Ray from source, installing specific versions, and more.
 ### Configuring REX-Ray
 REX-Ray requires a configuration file for storing details used to communicate
 with storage providers. This can include authentication credentials and
-driver-specific configuration options. Refer to the libStorage Storage Providers
-[documentation](http://libstorage.readthedocs.io/en/stable/user-guide/storage-providers/)
-for sample configurations of all supported storage platforms. Additionally, look
-at [core properties](./user-guide/config.md#configuration-properties) &
-[logging](./user-guide/config.md#logging-configuration) for advanced
-configurations.
+driver-specific configuration options. 
 
-Create a configuration file on the host at `/etc/rexray/config.yml`. Here is a
-simple example for using Oracle VirtualBox:
+After REX-Ray has been installed, copy and paste the contents from the [REX-Ray Configuration Generator](https://rexrayconfig.emccode.com/) to a new file on the host at `/etc/rexray/config.yml`. The configuration generator will satisfy the majority of use cases for REX-Ray.
+
+For more advanced and manual options, such as using multiple storage platforms, refer to the libStorage Storage Providers [documentation](http://libstorage.readthedocs.io/en/stable/user-guide/storage-providers/) and [core properties](./user-guide/config.md#configuration-properties) & [logging](./user-guide/config.md#logging-configuration).
+
+##### Example
+Here is a simple example for using Oracle VirtualBox:
 
 ```yaml
 libstorage:
   service: virtualbox
+virtualbox:
+  volumePath: $HOME/VirtualBox/Volumes
 ```
-
-Refer to the
-[VirtualBox documentation](http://libstorage.readthedocs.io/en/stable/user-guide/storage-providers/#virtualbox)
-for additional VirtualBox configuration options.
 
 From here, REX-Ray can now be used as a command line tool. View the commands
 available:
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -9,6 +9,12 @@ This page reviews how to configure REX-Ray to suit any environment, beginning
 with the the most common use cases, exploring recommended guidelines, and
 finally, delving into the details of more advanced settings.
 
+## Quick Configuration
+Utilize the [REX-Ray Configuration Generator](https://rexrayconfig.emccode.com/) to dynamically create a configuration file. The final configuration will be generated based upon the inputs for the supported storage platform. 
+
+After installing REX-Ray, create a configuration file on the host at `/etc/rexray/config.yml` and copy the contents from the [REX-Ray Configuration Generator](https://rexrayconfig.emccode.com/) using `vi`.
+
+Start REX-Ray as a Service with `$ rexray start`
 
 ## Basic Configuration
 This section outlines the two most common configuration scenarios encountered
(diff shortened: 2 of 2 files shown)


akutz:
Hi @kacole2,

I thought I commented on this earlier, but my comment is missing. Oops. I guess I neglected to hit `enter` before closing the tab.

Beyond the general "thank you" for doing this, I have some thoughts:

- The `rexray.logLevel` property should be removed as it is a default property and value.
- The default libStorage log level is also `warn`, and I think that property should be removed as well. The property path specified in the example is incorrect as is. The correct property path can be found in the libStorage [documentation](http://libstorage.readthedocs.io/en/stable/user-guide/config/#configuration-properties).
- I do not care for the new text under the "Create the File" heading. It's too crowded. I'd leave the generator text and relocate the additional explanation to some other section such as Advanced Configuration. 

However, beyond the above comments, I'm not sure we should have a generator at all until we can provide an out-of-the-box configuration that works without any additional configuration, per @cduchesne's comments he made to me via our Slack channel.

Without a VFS driver implementation we cannot provide a configuration example that works without additional, manual configuration on the host OS and potentially storage platform. Since removing the default log levels above results in **only** storage platform configuration properties in the generator, I do not think the generator provides that much value...**at this time.** 

Again, once we have a proper VFS implementation, then perhaps. Even then though, an example will simply be:

```
libstorage:
  service: vfs
```

So I'm not too sure we even need a generator, as cool as it is.

kacole2:
a few responses.

- What do you mean by `as it is a default property and value` regarding REX-Ray logging? The drop down gives you the choice of setting the logging level so you can stick with Warn as the default or change it. 
- `The default libStorage log level is also warn, and I think that property should be removed as well`, why remove it? The idea is to give you the ability to dynamically set the logging to whatever you want. If they want to keep it at Warn, then just leave it at Warn
- I can fix the property path for the logging level of libStorage. I missed that.

I get where we are going as an insider. But from the outside it isn't as clear and it's going to become more unclear when the masses start making their way to v0.4.X. I also missed @cduchesne comments about providing an out-of-the-box configuration for every supported storage platform. That would be great but I can't find the conversation.

I hope that a configuration generator wouldn't be necessary, and maybe it won't in the coming months, but the amount of options needed for configuring different storage platforms (Isilon, ScaleIO, AWS, VirtualBox, OpenStack) leaves plenty of room for making things extremely simplified for the end-user. Thus, the generator can be used as a band-aid for you not having to read the next hundred or so issues that are all being driven from incorrect configuration settings.

akutz:
Hi @kacole2,

Regarding your first and second responses - I do not care for specifying properties that are set via their default value. Not setting `libstorage.logging.logLevel` for instance still means that property is set to `warn`. What I mean by it showing up by default is that it is in the produced example whether or not a user has selected a value other than `warn`.

This isn't my comment as an insider. I used to think the same as you until @cduchesne remarked about the way we start the service by default when installing REX-Ray. I said so does Apache, Tomcat, PostgreSQL, etc. He rightly responded that those services all provide working configurations out-of-the-box. We can't. Not that we don't, but we can't. Simply because there is currently no method of deploying REX-Ray with libStorage without a dependency on some external storage driver.

My plan is to finish the VFS driver and enable an out-of-the-box configuration scenario I described above. At that point a configuration generator would make a lot of sense as it could generate a default configuration that would at least show how to configure VFS. Since VFS would be pure Go, embedded in libStorage, a user could take the configuration, drop it in, and start it up without any additional work on their part.

Does that make sense? 

Also, regarding the generator, you can define more than one service. That's part of the basic value of libStorage. How difficult would it be to change that? You can see examples of this in the libStorage and REX-Ray documentation.

akutz:
Also, I just noticed the RTFD link in the upper right-hand corner points to the `latest` docs. It should point to `stable`:

http://rexray.readthedocs.io/en/stable/

akutz:
One more thing...remove the libStorage logging level. I am coming around to your view on this @kacole2. To that end, remove the libStorage logging level altogether. I automatically pick up the RR log level and assign it to LS pretty early in the start up. No reason to have both there. I'll re-review this from a simple perspective as you have outlined. I see your intent more clearly now I think.

akutz:
Also, the web app itself. Where is it? I may want to PR it. I like it, but it should remove unused config branches if a property is unchecked. And it should also have tool tips or links to property explanations beside the property checkboxes.

kacole2:
@akutz it's located here (https://github.com/kacole2/rexray-config-generator) Send in all your feature requests and I can work on them. I totally agree with your points about it needing to be removed if unchecked as well as adding tooltips for individual pieces. This was just the fastest way for me to get it out. All of that is on my TODO

kacole2:
@akutz sort of "get it out there and see what happens" instead of building for every use case at the first go. But please don't waste your time on it. You've got bigger fish to fry and I can try to make this fit into your vision or we can just toss it in the garbage now if we want :)

akutz:
Hi @kacole2,

No worries. I think I do like it actually. Let's maybe add the caveat that this doesn't mean RR will actually start. It's just a way to graphically edit the config with known options. To that end it would be nifty to actually have the tooltips and explanations I mentioned above. 

Also, no coffee script? No bower? :(

akutz:
Hmm. I'm wondering if there is a way to express the config options via a JSON schema you could then read to produce your GUI? I know that there are other tools that do that:

- http://schemaform.io
- http://jsonschema.net

kacole2:
@akutz i'm going to start putting issues into it now so i can track them. I can have tooltips in there pretty quickly. 

no coffee script. all Angular.js. Haven't used Bower before. i guess this is a learning opportunity! 

funny enough, after I got done with this and was trying to implement some tests I realized that i should have used Yeoman from the beginning.... next time

akutz:
Yup, yeoman is my skeleton provider of choice. I wrote all of EMC's Big Data Lake UI stuff with a Yeoman+Bower+AngularJS+CoffeeScript app :)

kacole2:
well hell... schemaform may have made my life a bit easier. didn't know that existed

akutz:
There is a handful of things like this. I've looked at a bunch of them for a similar reason. Google "json schema viewer" or editor or variations of that. Lots of them are pure JS too :)

kacole2:
Added tooltips https://github.com/kacole2/rexray-config-generator/pull/11

kacole2:
Added filter to remove pieces that are blank or not set to a true value. https://github.com/kacole2/rexray-config-generator/pull/13

kacole2:
changed logging levels to be independent as you suggested. https://github.com/kacole2/rexray-config-generator/pull/17

kacole2:
changes to the docs were made as you requested @akutz. Rebased and ready for merge. Let me know if there is something else you would like to see

akutz:
Hi @codenrhoden,

The reason this has never been merged is because the original file went into conflict very early past its submission. If you or @kacole2 want to update the PR we can merge it into the 0.9.1 release branch.

codenrhoden:
@kacole2 Can you please rebase this so we can get it merged in this week?

codenrhoden:
Replaced by #888

I had to replace the PR because this one has an origin of "unknown repo", so I had no way to keep/merge the original commit and fix it in a merge commit. Weird. Found several other things to change as well. See #888 for details.

PR #505, opened 2016-07-13
Fix for Profiler Race Condition

akutz:
The profiler added to REX-Ray worked fine until it was relocated to the `init()` function from the `main()` function. This was done in hopes of capturing the effects of loading the other packages. However, it ended up stalling the application randomly, likely because of a race condition created trying to load profiling when the profiling packages were not yet loaded themselves.

This patch moves the profiler code back into the `main()` function.

This patch also makes the program behave differently upon being asked to close via a signal. If the application is running as a foreground service and Ctrl-C is used to quit, that actually sends the SIGINT signal. This will allow for a graceful shutdown and a successful exit code. However, using the `kill` command normally or with `kill -9` causes the program to abort immediately with a non-zero exit code.

Signal | Graceful Shutdown | Exit Code | Method
-------|--------------------|-----------|-------- 
SIGKILL | no | 1 | `kill -KILL PID`
SIGTERM | no | 1 | `kill PID` (or `kill -TERM PID`)
SIGHUP | yes | 0 | `kill -HUP PID`
SIGINT | yes | 0 | `kill -INT PID` (if the service is running in the foreground, using `Ctrl-C` to stop the service actually sends the process this signal)
SIGQUIT | yes | 0 | `kill -QUIT PID` (if the service is running in the foreground, using `Ctrl-\` to stop the service actually sends the process this signal)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -470,7 +470,8 @@ import (
 
 func init() {
 	Version = &apitypes.VersionInfo{}
-	Version.Arch = fmt.Sprintf("%s-%s", os(runtime.GOOS), arch(runtime.GOARCH))
+	Version.Arch = fmt.Sprintf(
+		"%s-%s", osString(runtime.GOOS), archString(runtime.GOARCH))
 	Version.Branch = "$(V_BRANCH)"
 	Version.BuildTimestamp = time.Unix($(V_EPOCH), 0)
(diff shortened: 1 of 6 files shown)


vladimirvivien:
@akutz the pprof addition looks great.

PR #513, opened 2016-07-20
doc: simple fixes

codenrhoden:
It had been a while since I did a full read-through of the docs. Caught a few minor things that were easy to fix.


akutz:
Hi @codenrhoden,

Thank you so much for this!

PR #514, opened 2016-07-22
Update badge links in README.md

cduchesne:
The current README doesn't show all badges correctly. This changes the links to https from http, which fixes the problem.


cduchesne:
closing for now as the http links are no longer failing

PR #540, opened 2016-08-25
Refine rexray volume ls filter

proudh:
This commit fixes the following issues:
1. When `rexray volume ls` was used specifying only a volume ID,
either a volume with a blank name, or a volume matching the same
volume ID would be returned.
2. When `rexray volume ls` was used specifying a volume name that
matched multiple volumes, only one volume with a matching volume
name would be returned. In REX-Ray 0.3.0, all volumes that matched
the volume name would be returned.

### DIFF
--- a/rexray/cli/cmds_volume.go
+++ b/rexray/cli/cmds_volume.go
@@ -66,14 +66,13 @@ func (c *CLI) initVolumeCmds() {
 			}
 			if c.volumeID != "" || c.volumeName != "" {
 				for _, v := range vols {
-					if strings.ToLower(v.ID) == strings.ToLower(c.volumeID) ||
-						strings.ToLower(v.Name) == strings.ToLower(c.volumeName) {
+					if (c.volumeID != "" && strings.EqualFold(v.ID, c.volumeID)) ||
+						(c.volumeName != "" && strings.EqualFold(v.Name, c.volumeName)) {
 						out, err := c.marshalOutput(v)
 						if err != nil {
 							log.Fatal(err)
 						}
 						fmt.Println(out)
-						return
 					}
 				}
 				return


rexray/cli/cmds_volume.go
```diff
@@ -66,14 +66,13 @@ func (c *CLI) initVolumeCmds() {
 			}
 			if c.volumeID != "" || c.volumeName != "" {
 				for _, v := range vols {
-					if strings.ToLower(v.ID) == strings.ToLower(c.volumeID) ||
-						strings.ToLower(v.Name) == strings.ToLower(c.volumeName) {
+					if (c.volumeID != "" && strings.ToLower(v.ID) == strings.ToLower(c.volumeID)) ||
```
akutz: Hi @proudh,

This looks good, thank you! Could you please do me a favor and update the logic? I recently read [an article](https://medium.com/@benbjohnson/go-walkthrough-bytes-strings-packages-499be9f4b5bd) by @benbjohnson on ways to optimize Go code, and string comparisons are one area that can be greatly improved. To achieve efficient, case-insensitive string comparison, please use [`string.EqualFold`](https://golang.org/pkg/strings/#EqualFold) instead. Memory is allocated with every `strings.ToLower` call, and this is avoided by using the `EqualFold` comparison function. I've started to use it wherever I would normally have used a `ToLower` or `ToUpper` to perform case-insensitive comparison. 

Thanks!

rexray/cli/cmds_volume.go
```diff
@@ -66,14 +66,13 @@ func (c *CLI) initVolumeCmds() {
...
+						(c.volumeName != "" && strings.ToLower(v.Name) == strings.ToLower(c.volumeName)) {
```
akiraw: (c.volumeName != "" && strings.EqualFold(v.Name, c.volumeName))

akutz: Hi @akiraw,

I'm unsure of the question you're posing here. Are you asking if the above code is the desired approach?

akiraw: Yes, we were wondering if the change you requested was complicated or if the code I posted above was sufficient. My apologies, I should have been more clear in my line comment.
Thank you for your help.

proudh:
Hi @akutz, nice find!  I've made the change for optimizing string comparison based on @benbjohnson's article.

cduchesne:
nice work @proudh !

Issue #609, opened 2016-10-26
EFS can't be mounted in 0.6.0


mhrabovcin:
Trying to mount EFS volume by its name doesn't work in `0.6.0`. Mounting works with version built manually before this commit - https://github.com/emccode/rexray/commit/c68e7d706e5167af9ffa5423b82325b8f47ba7ea

mhrabovcin:
Rexray version:

```
[root@ip-10-0-0-4 ~]# ./rexray version            
REX-Ray                                           
-------                                           
Binary: /root/rexray                              
SemVer: 0.6.0                                     
OsArch: Linux-x86_64                              
Branch: v0.6.0                                    
Commit: 429cfdf7272cebc5911b2865128ad37a6da5a203  
Formed: Fri, 21 Oct 2016 02:06:45 UTC             

libStorage                                        
----------                                        
SemVer: 0.3.1                                     
OsArch: Linux-x86_64                              
Branch: v0.6.0                                    
Commit: a3a561a1b94bb8b7efb4ae998540c457deb39692  
Formed: Fri, 21 Oct 2016 02:05:09 UTC
```

kasisnu:
Some more debugging info:

```[root@ip-10-0-0-4 ~]# ./rexray version
REX-Ray
-------
Binary: /root/rexray
SemVer: 0.6.0
OsArch: Linux-x86_64
Branch: v0.6.0
Commit: 429cfdf7272cebc5911b2865128ad37a6da5a203
Formed: Fri, 21 Oct 2016 02:06:45 UTC

libStorage
----------
SemVer: 0.3.1
OsArch: Linux-x86_64
Branch: v0.6.0
Commit: a3a561a1b94bb8b7efb4ae998540c457deb39692
Formed: Fri, 21 Oct 2016 02:05:09 UTC
 
[root@ip-10-0-0-4 ~]# ./rexray volume ls
ID           Name       Status     Size
fs-53ad791a  wutututut  available  6144


[root@ip-10-0-0-4 ~]# ./rexray volume mount wutututut --service efs
FATA[0000] no volumes found

```

With http response logging enabled, we've confirmed volumes are being returned by the libstorage client

```e=objectivefs storageDriver=libstorage time=1477475869228 txCR=1477475869 txID=80697ee2-949c-4749-7d12-85b797132905
[...]

PR #615, opened 2016-10-31
Revert "Fix for No Volumes Avai for some commmands"

akutz:
Reverts codedellemc/rexray#613

This patch is being reverted because it breaks some drivers, such as EBS, which return only volumes that are attached to the client when this value is set.

Further discussion of #609 is required.

### DIFF
--- a/cli/cli/cmds_volume.go
+++ b/cli/cli/cmds_volume.go
@@ -803,12 +803,6 @@ func (c *CLI) lsVolumes(
 	args []string,
 	validStatuses ...string) (*lsVolumesResult, error) {
 
-	// if the volume status matters then send the flag that requests a
-	// volume's attachment information
-	if len(validStatuses) > 0 {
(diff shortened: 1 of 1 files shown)


akutz:
Hmm, I'm closing this for now as I'm not sure it's not libStorage or the EBS driver in error. Further research is required.

akutz:
Okay, it appears the libStorage framework is [removing volumes](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L182) that do not have attachments when the `attachments` flag is provided. This is not driver-specific.

@clintonskitson, I believe I had instrumented this at your request. This was where we needed the bitmask to describe the different behaviors. I am going to disable this filtering until such a time the bitmask is implemented (if not now) as it is preventing some functionality in REX-Ray from working.

(cc @mhrabovcin, @kasisnu, @cduchesne)

PR #619, opened 2016-11-04
libStorage 0.3.3-rc3

akutz:
This patch updates REX-Ray to work with libStorage 0.3.3-rc3 and its new volume attachment query/filter framework.

### DIFF
--- a/cli/cli/cmds_volume.go
+++ b/cli/cli/cmds_volume.go
@@ -10,7 +10,6 @@ import (
 	"github.com/spf13/cobra"
 
 	apitypes "github.com/codedellemc/libstorage/api/types"
-	apiutils "github.com/codedellemc/libstorage/api/utils"
 )
 
 func (c *CLI) initVolumeCmdsAndFlags() {
@@ -36,7 +35,7 @@ func (c *CLI) initVolumeCmds() {
 		Short:   "List volumes",
 		Example: "rexray volume ls [OPTIONS] [VOLUME...]",
 		Run: func(cmd *cobra.Command, args []string) {
-			c.mustMarshalOutput(c.lsVolumes(args))
+			c.mustMarshalOutput(c.lsVolumes(args, c.defaultAttachments()))
 		},
 	}
 	c.volumeCmd.AddCommand(c.volumeListCmd)
@@ -179,7 +178,7 @@ func (c *CLI) initVolumeCmds() {
 		Run: func(cmd *cobra.Command, args []string) {
 			checkVolumeArgs(cmd, args)
 
-			result, err := c.lsVolumes(args)
+			result, err := c.lsVolumes(args, requestAttachments())
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -229,7 +228,9 @@ func (c *CLI) initVolumeCmds() {
 		Run: func(cmd *cobra.Command, args []string) {
 			checkVolumeArgs(cmd, args)
 
-			result, err := c.lsVolumes(args, volumeStatusAvailable)
+			result, err := c.lsVolumes(
+				args,
+				requestAttachments(apitypes.VolumeAttachmentsUnattached))
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -282,7 +283,11 @@ func (c *CLI) initVolumeCmds() {
 		Run: func(cmd *cobra.Command, args []string) {
 			checkVolumeArgs(cmd, args)
 
-			result, err := c.lsVolumes(args, volumeStatusAttached)
+			result, err := c.lsVolumes(
+				args,
+				requestAttachments(
+					apitypes.VolumeAttachmentsMine,
+					apitypes.VolumeAttachmentsAttached))
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -339,7 +344,11 @@ func (c *CLI) initVolumeCmds() {
 			// volumes that are attached or can be attached should be valid
 			// candidates for a mount operation
 			result, err := c.lsVolumes(
-				args, volumeStatusAttached, volumeStatusAvailable)
+				args,
+				requestAttachments(
+					apitypes.VolumeAttachmentsMine,
+					apitypes.VolumeAttachmentsAttached,
+					apitypes.VolumeAttachmentsUnattached))
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -396,7 +405,11 @@ func (c *CLI) initVolumeCmds() {
 		Run: func(cmd *cobra.Command, args []string) {
 			checkVolumeArgs(cmd, args)
 
-			result, err := c.lsVolumes(args, volumeStatusAttached)
+			result, err := c.lsVolumes(
+				args,
+				requestAttachments(
+					apitypes.VolumeAttachmentsMine,
+					apitypes.VolumeAttachmentsAttached))
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -445,7 +458,11 @@ func (c *CLI) initVolumeCmds() {
 		Short:   "Print the volume path",
 		Example: "rexray volume path [OPTIONS] [VOLUME...]",
 		Run: func(cmd *cobra.Command, args []string) {
-			result, err := c.lsVolumes(args, volumeStatusAttached)
+			result, err := c.lsVolumes(
+				args,
+				requestAttachments(
+					apitypes.VolumeAttachmentsMine,
+					apitypes.VolumeAttachmentsAttached))
 			if err != nil {
 				log.Fatal(err)
 			}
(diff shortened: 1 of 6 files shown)


cli/cli/cmds_volume.go
```diff
@@ -179,7 +178,7 @@ func (c *CLI) initVolumeCmds() {
 		Run: func(cmd *cobra.Command, args []string) {
 			checkVolumeArgs(cmd, args)
 
-			result, err := c.lsVolumes(args)
+			result, err := c.lsVolumes(args, requestAttachments())
```
codenrhoden: Since we are requesting attachments here, should we be using that info to see if the volume is attached before doing the remove?

cli/cli/cmds_volume.go
```diff
@@ -339,7 +344,11 @@ func (c *CLI) initVolumeCmds() {
 			// volumes that are attached or can be attached should be valid
 			// candidates for a mount operation
 			result, err := c.lsVolumes(
-				args, volumeStatusAttached, volumeStatusAvailable)
+				args,
+				requestAttachments(
+					apitypes.VolumeAttachmentsMine,
+					apitypes.VolumeAttachmentsAttached,
```
codenrhoden: Should `VolumeAttachmentsAttached` be here? That would be volumes that are attached to other hosts, which I believe would be unavailable for mounting...

Is this need for pre-emption? (I'm assuming so)

cli/cli/cmds_volume.go
```diff
@@ -339,7 +344,11 @@ func (c *CLI) initVolumeCmds() {
...
+					apitypes.VolumeAttachmentsUnattached))
```
codenrhoden: sorry to create a new review, but I hit a problem with testing today. I'm not sure if the solution is a change to rex, or to libstorage, but as it stands with LS 0.3.3-rc3, you can't request both attached and unattached volumes at the same time.

this is because if you specific attached volumes, any volumes that are unattached are filtered out: https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L181-L187

The inverse turns out to be true, but the check for attached happens before checking for unattached.

akutz:
Ah yes. I have this fixed on a PR not yet merged. It checks for the different states and handles them appropriately. Got side tracked with minimal client.

-- 
-a

cli/cli/cmds_volume.go
codenrhoden: @akutz.  I think this still has an issue. The block for `attachements.Mine` is filtering `vol.Attachments`, removing anything that does not have a matching IID completely. This makes sense on its own, but gives `Mine` precedence over `Attached` and `Unattached`.

For example,  the REX-Ray code that I commented on in this thread uses:
```
requestAttachments(
					apitypes.VolumeAttachmentsMine,
					apitypes.VolumeAttachmentsAttached,
					apitypes.VolumeAttachmentsUnattached))
```

The use of `Mine` with `Unattached` is still not going to get what it's asking for.

That being said, I'm still not sure that combination of masks makes sense. It essentially says "give me everything that's attached to me, everything that's attached at all (including any other hosts), and anything that is unattached". Doesn't that mean "give me everything"? Why use the bitmask at all, then, instead of just `Requested`?

Put another way, `Mine` still cannot be used with `attached` or `detached`, *if* the intent of `attached` and `detached` is to show you volumes that are not yours.

Using `Mine` + `Attached` + `Unattached` seems to have no purpose -- just use `Requested`
Using `Attached` + `Detached` in any scenario has no purpose -- at that point you get everything and should just use `Requested`.
Combining `Mine` with `Attached` seems misguided. `Mine` will always be a subset of `Attached`, and is thus pointless unless you submit `Mine` on its own.
Combining `Mine` with `Unattached` has potential. Basically, show me what I have and show what is available. But the current implementation would block that from working.

at least that's how I'm reading it.

akutz: Hi @codenrhoden,

So I hear you and I understand. I think part of the issue is that I want to be able to do this, that, and the other with a single query. Meaning I'd like to be able to say "get all volumes and their attachments" or "get all volumes that are attached" or "get all volumes that are not attached" or "get all volumes that are attached to me and all volumes that are not attached".

What I don't want is for people to have to worry about is performing multiple queries to get what they want. Let's review the values here:

```
const (
	// VolumeAttachmentsRequested indicates attachment information is requested.
	VolumeAttachmentsRequested VolumeAttachmentsTypes = 1 << iota // 1

	// VolumeAttachmentsMine indicates attachment information should
	// be returned for volumes attached to the instance specified in the
	// instance ID request header. If this bit is set then the instance ID
	// header is required.
	VolumeAttachmentsMine // 2

	// VolumeAttachmentsDevices indicates an attempt should made to map devices
	// provided via the local devices request header to the appropriate
	// attachment information. If this bit is set then the instance ID and
	// local device headers are required.
	VolumeAttachmentsDevices // 4

	// VolumeAttachmentsAttached indicates only volumes that are attached
	// should be returned.
	VolumeAttachmentsAttached // 8

	// VolumeAttachmentsUnattached indicates only volumes that are unattached
	// should be returned.
	VolumeAttachmentsUnattached // 16
)
```

As I see it, here are the valid masks:

Mask | Components | Description
------|--------------|------------
`0`| `0` | The default value. This indicates no attachment information is requested.
`1` | `1` | Request attachment information for all retrieved volumes
`3` | `1|2` | Request attachment information for volumes attached to the instance provided in the instance ID
`7` | `1|2|4` | Request attachment information for volumes attached to the instance provided in the instance ID and perform device mappings where possible.
`9` | `1|8` | Request attachment information for all retrieved volumes and return only volumes that are attached to some instance.
`17` | `1|16` | Request attachment information for all retrieved volumes and return only volumes that are not attached to any instance.

Here's where it gets a little more interesting:

Mask | Components | Description
------|--------------|------------
`11` | `1|2|8` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID.
`27` | `1|2|8|16` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID *or* are not attached to any instance at all. tl;dr - _Attached To Me or Available_

akutz: Hi @codenrhoden,

I got cut off above (hit enter too soon). I may be totally off base above, but can you see what I was attempting with the above mask?

One thing to note, and in regards to something you said:

> Why use the bitmask at all, then, instead of just `Requested`?

Because to do any of the mapping to an instance or device mapping requires that clients send in the request headers the instance ID and local device information. The mask provides the ability to retrieve volumes without also requiring the presence of the instance ID and/or local device information in the headers.

Previously I assumed drivers would check to see if the instance ID was present, and if so, automatically return volume attachment information for only that instance. 

However that was too constricting as it didn't leave us the ability to create a client that automatically handles sending the instance ID for a user. Additionally, it meant for some drivers we could not request volumes that are both attached to a provided instance but also available (unattached to any other instance). 

See what I mean?

Or am I totally missing something obvious here? I could be. What are your thoughts?

FWIW, see [these tests](https://github.com/akutz/libstorage/blame/bugfix/attachment-filtering-logic/drivers/storage/vfs/tests/vfs_test.go#L166-L312) I added to showcase the various ways to use the mask. Each test asserts the return values are what they are expected to be.

akutz: Hi @codenrhoden,

I just noticed this after re-reading your comments:

> This makes sense on its own, but gives Mine precedence over Attached and Unattached.

Well, `mine` has no precedence over `unattached` as `mine` only has any effect if a volume has attachments.

And yes, `mine` has precedence over `attached`, but if you take a close look at the tables above you'll notice there is no attempt of the following:

> Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID or are not attached to any instance at all or **are attached at all**.

The above would be what you described -- a mask that's the same as saying get all attachments for all volumes, right?

As it is, the logic is such to enable requesting all volumes attached to the provided instance or all volumes not attached to any instance, aka. volumes available to be attached.

The order of the server logic is specifically designed to enable this. That's why the `attached` bit only removes unattached volumes if the `unattached` bit is not set. That's why the `unattached` bit only removed attached volumes if the`attached` bit is not set.

Again, I do have [these tests](https://github.com/akutz/libstorage/blob/bugfix/attachment-filtering-logic/drivers/storage/vfs/tests/vfs_test.go#L166-L312) that validate the desired logic. 

I'm fully willing to admit I'm off base on all of this, but I also think it's possible I just did a **supremely** shitty job describing the purpose and implementation of these things with which to begin, and it's lead to much confusion.

codenrhoden: @akutz 

>I'm fully willing to admit I'm off base on all of this, but I also think it's possible I just did a supremely shitty job describing the purpose and implementation of these things with which to begin, and it's lead to much confusion.

No no, I totally understand the purpose, and think the implementation is great. I just narrowed in on a usecase where I thought the implementation would return invalid results. After going through all your comments, I *think* I've identified the crux of it all.

We have different interpretations of mask `1|2|8`.  You defined this as:
>Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID.

I would have defined it as:
Request attachment information for all retrieved volumes and return volumes that are attached to the instance provided in the instance ID **or attached to a different instance**.

I think `1|2|8` is functionally equivalent to `1|8`, and you think it is equivalent to `1|2`.  From your table:

| Mask | Compnents | Description |
| ---- | ------------ | ----------- |
| `3` | `1|2` | Request attachment information for volumes attached to the instance provided in the instance ID |
| `11` | `1|2|8` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID. |

I do not see a difference between those two. Do you? Is there a scenario where those two masks would return different results?

This is why I was saying `Mine` has precedence over `Attached`. In my scenario, it was going to keep volumes attached to other instances from being returned -- I thought that was wrong. You are saying that is by design.

This is totally fine! I just think we need to make that clear. My confusion stemmed from not seeing why you want to use mask `1|2|8` unless you *wanted* to get volumes attached to other instances, because otherwise you would just use mask `1|2`.

Does that make sense?

akutz: It does make sense. However, the masks `3` and `11` are **intended** to return different results.

A mask of `3` should return **all** volumes, but only bother fetching/returning the attachment information for the instance specified in the instance ID header.

A mask of `11` should return **only** the volumes attached to the instance specified in the instance ID header. I worded the description poorly as technically a driver doesn't have to retrieve info for all volumes if the backend has an API that allows it to return attachment info for a particular instance. However, I was thinking along the lines of how you likely have to implement it -- get all volumes and their attachment info, filter out only the volumes that are attached to a provided instance.

Essentially the **intended** difference between `3` and `11` is that in `3` all volumes are returned, but only the ones attached to the provided instance ID have attachment info. For mask `11` only the volumes attached to the provided instance ID have attachment info and are the only ones returned.

Let's look at the code to which you refer:

```go
// if only the requesting instance's attachments are requested then
	// filter the volume's attachments list
	if attachments.Mine() {
		atts := []*types.VolumeAttachment{}
		for _, a := range vol.Attachments {
			alf := log.Fields{
				"attDeviceName": a.DeviceName,
				"attDountPoint": a.MountPoint,
				"attVolumeID":   a.VolumeID,
			}
			if strings.EqualFold(iid.ID, a.InstanceID.ID) {
				atts = append(atts, a)
				ctx.WithFields(lf).WithFields(alf).Debug(
					"including volume attachment")
			} else {
				ctx.WithFields(lf).WithFields(alf).Debug(
					"omitting volume attachment")
			}
		}
		vol.Attachments = atts
		ctx.WithFields(lf).Debug("included volume attached to instance")
	}
```

The interior of the `Mine()` scope doesn't omit the **volume** -- only its attachment(s). If the `2` bit is set, only attachments attached to the provided instance are retained in the attachments list. That's separate and unrelated to whether a volume is returned based on the presence of bits `8` or `16`:

```go
	// if the volume has no attachments and the mask indicates that
	// only attached volumes should be returned then omit this volume
	if len(vol.Attachments) == 0 &&
		attachments.Attached() &&
		!attachments.Unattached() {
		ctx.WithFields(lf).Debug("omitting unattached volume")
		return false
	}

	// if the volume has attachments and the mask indicates that
	// only unattached volumes should be returned then omit this volume
	if len(vol.Attachments) > 0 &&
		!attachments.Attached() &&
		attachments.Unattached() {
		ctx.WithFields(lf).Debug("omitting attached volume")
		return false
	}

	return true
```

Or am I still missing something here?

codenrhoden: > Or am I still missing something here?
Nope - I finally get it. I'm dense. :)

I should have been able to see the difference, but just couldn't. Thanks for sticking with me and not just writing me off.

akutz:
Closing in favor or #622.

Issue #629, opened 2016-11-22
Populate deviceName where applicable in volume commands


cduchesne:
Is it possible to populate the deviceName field in volume commands, such as ls/get/inspect?

```
# rexray volume ls --format jsonp
[
  {
    "attachments": [
      {
        "deviceName": "",
        "instanceID": {
          "id": "c00e3a6200000008",
          "driver": "scaleio"
        },
        "status": "",
        "volumeID": "925def4300000000"
      }
    ],
    "availabilityZone": "default",
    "name": "test",
    "size": 16,
    "id": "925def4300000000",
    "type": "default"
  }
]
```

codenrhoden:
@cduchesne Great question! :)  Add the `--path` flag and you will find what you seek.

PR #630, opened 2016-11-22
Always request device mapping for vol ls commands

codenrhoden:
@cduchesne I think a change like this is what you'd be looking for in #629.

@akutz will have to weigh in on whether it is appropriate to do so. A requirement for getting the Device map info is the client has to send both its Instance ID and its LocalDevice headers. A lot of the changes that have been happing in libStorage 0.3.3 are to account for the fact that drivers should not always expect those to be present. They should only be present when they are required (like for doing a device mapping).

I'm not sure right now if rexray, as a client, is always sending those are not.

The change for `c.volumeAttachedToMe` is probably a good one. This correlates to `rexray volume ls --attachedToMe`.

But when getting all attached vols `--attached`, or *all vols*, I'm not sure if rexray is sending those headers or not. There also aren't convenience fields created with the masks we needed (why I'm `or`'ing two masks) which is a sign to me that maybe that combination is not intended.

But, I always like to throw something out to see what is possible, for discussions. :)

### DIFF
--- a/cli/cli/cmds_volume.go
+++ b/cli/cli/cmds_volume.go
@@ -812,12 +812,14 @@ func (c *CLI) defaultAttachments() apitypes.VolumeAttachmentsTypes {
 		return apitypes.VolAttReqOnlyUnattachedVols
 	}
 	if c.volumeAttachedToMe {
-		return apitypes.VolAttReqOnlyVolsAttachedToInstance
+		return apitypes.VolAttReqWithDevMapOnlyVolsAttachedToInstance
 	}
 	if c.volumeAttached {
-		return apitypes.VolAttReqOnlyAttachedVols
+		return apitypes.VolAttReqOnlyAttachedVols |
+			apitypes.VolumeAttachmentsDevices
 	}
-	return apitypes.VolAttReq
+	return apitypes.VolAttReq |
+		apitypes.VolumeAttachmentsDevices
 }
 
 func (c *CLI) lsVolumes(


akutz:
Hi @codenrhoden (cc @cduchesne),

So the reason we don't do this already is because of, like Travis said, the penalty incurred when asking that local device information be sent / correlated every time. It's one reason the `rexray volume path` command is separate from `rexray volume ls`. By default the `ls` command doesn't perform the device mapping, leaving that explicitly to the `path` command.

I don't actually like how I've handled the CLI flags. I'm not sure a mask from the CLI perspective is much better. Honestly I think the right way to do this may be to remove the changes in this PR, the flags that I've previously defined, and do the following:

1. Remove the `volume path` command.
2. Make `--path` a flag for the `volume ls` command, and if specified, path information is retrieved during an `ls` operation.
3. Create a new flag called `--filter` and let its options be `attached` and `available`. Based on which filter is applied, we apply the bit that tells the server to return either only attached volumes or available ones.

Thoughts?

codenrhoden:
>1. Remove the volume path command.
>2. Make --path a flag for the volume ls command, and if specified, path information is retrieved during an ls operation.

I like this

>3. Create a new flag called --filter and let its options be attached and available. Based on which filter is applied, we apply the bit that tells the server to return either only attached volumes or available ones.

Would there also be a filter for `attachedToMe` ?

In general I like the approach. Those filters don't necessarily have to include the device mappings, though.  Should there be an additional flag, `--devices` that sets that bit as well?  That way it is up to the user to say whether they want device info returned.

akutz:
Hi @codenrhoden,

Instead of `attachedToMe` I think I'd add a flag called `--mine` that implies `attached` as well. That or filter based on `attached` (attached to me), `unavailable` (attached to someone else), or `available`.

Those flags would *not* imply device mappings. That's the point of the `--path` flag, right? Since path correlation requires device mapping information.

codenrhoden:
Okay, that makes sense. So if you wanted to get full details of your attache volumes, you would do:

`rexray volume ls --mine --path`

Correct?

akutz:
Hi @codenrhoden,

Unless we wanted it to be `rexray volume ls --filter attached --path`. I kind of like the filter options I specified above where `attached` means _mine_ and `unavailable` means _attached to others_

akutz:
Or instead of `--filter attached|available|unavailable` it could be `--attached`, `--unavailable`, `--available`...

codenrhoden:
I would use `--filter` **IF** only one can be applied. Otherwise, use individual flags. I personally dislike having to check if `--flag1` and `--flag2` are incompatible with each other. Sometimes can't be avoided, though.

And I do agree with you that `--attached` should already mean *mine*.

my two cents.

codenrhoden:
@akutz Once you decide on a path forward, feel free to close this PR.

akutz:
Well, what do you think then? This is what it feels like we just concluded:

1. Add a `--path` flag for the `volume ls` command.
2. Add the flags `--attached`, `--available`, and `--unavailable`. 

Does that make sense to you?

codenrhoden:
Yes I believe it does. Onward!

akutz:
Closed in favor of PR #634.

PR #641, opened 2016-12-01
Add bash script for k8s flex driver

codenrhoden:
This introduces a 'flexrex' script that can be installed and used as a Kubernetes
Flex Volume driver. See: https://github.com/kubernetes/kubernetes/tree/master/examples/volumes/flexvolume

This script makes calls to rexray for attach/detach, but does not use rexray for
mount/umount because the behavior that k8s needs is different from what
is built into the Docker integration driver.

Examples:

```
rexray volume create test --size 5
flexrex attach '{"volumeID":"test"}'
flexrex mount '/mnt/test' '/dev/sdb' '{"kubernetes.io/fsType":"ext4"}'
flexrex unmount /mnt/test
flexrex detach '/dev/sdb'
```

I made sure that both xfs and ext4 fs types work. The lvm example given for the flex driver does not work on unpartitioned block devices.

Testing has not been done with k8s proper at this point, only with the `flexrex` script directly, as described in the examples above.  cc/ @cduchesne 

This script only works with rexray starting with v0.6.1, as it requires the `volume ls --path` functionality`.  Not entirely sure where the script should live at this point, so I just proposed a place. Purpose of PR is to get to @cduchesne for testing.

### DIFF
--- a/cli/k8sflex/flexrex
+++ b/cli/k8sflex/flexrex
@@ -1,0 +1,190 @@
+#!/bin/bash
+
+# Copyright 2016 Dell EMC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Notes:
+#  - Please install "jq" package before using this driver.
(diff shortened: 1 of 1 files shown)


cli/k8sflex/flexrex
```diff
@@ -0,0 +1,190 @@
...
+}
+
+attach() {
+	VOLUMEID=$(echo $1 | jq -r '.volumeID')
+	if [ -z $VOLUMEID ]; then
+		err "{\"status\": \"Failure\", \"message\": \"Unable to extract volumeID\"}"
+		exit 1
+	fi
+
+	OUTPUT=$(rexray volume attach ${VOLUMEID} --format json)
```
codenrhoden: I guess I don't need to store the command results here. Originally I was using the output to look for the device, but I have to make a second call to get it.  device info doesn't come back from the attach command -- it can't because that relies on the localdevice headers, which are gathered/sent *before* the attach (obviously). so no mapping can be done.

Issue #589, opened 2016-10-09
Detect running libstorage service socket file and delete


cduchesne:
When running REX-Ray using a custom sock file path, REX-Ray will not remove / re-use the existing sock file and will crash if it exists already.

```
[root@host ~]# rexray status
● rexray.service - rexray
   Loaded: loaded (/etc/systemd/system/rexray.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Sat 2016-10-08 21:23:01 PDT; 3s ago
  Process: 21340 ExecStart=/usr/bin/rexray start -f (code=exited, status=2)
 Main PID: 21340 (code=exited, status=2)

Oct 08 21:23:01 mesos rexray[21340]: github.com/emccode/rexray/vendor/github.com/spf13/cobra.(*Command).Execute(0xc42013c000, 0xdf8038, 0xc4204c5d68)
Oct 08 21:23:01 mesos rexray[21340]: /home/travis/gopath/src/github.com/emccode/rexray/vendor/github.com/spf13/cobra/command.go:617 +0x398
Oct 08 21:23:01 mesos rexray[21340]: github.com/emccode/rexray/rexray/cli.(*CLI).execute(0xc4200fea80)
Oct 08 21:23:01 mesos rexray[21340]: /home/travis/gopath/src/github.com/emccode/rexray/rexray/cli/cli.go:265 +0x4b
Oct 08 21:23:01 mesos rexray[21340]: github.com/emccode/rexray/rexray/cli.(*CLI).Execute(0xc4200fea80)
Oct 08 21:23:01 mesos rexray[21340]: /home/travis/gopath/src/github.com/emccode/rexray/rexray/cli/cli.go:248 +0x71
Oct 08 21:23:01 mesos rexray[21340]: github.com/emccode/rexray/rexray/cli.Execute(0x12ad5a0, 0xc420120b40)
Oct 08 21:23:01 mesos rexray[21340]: /home/travis/gopath/src/github.com/emccode/rexray/rexray/cli/cli.go:215 +0x43
Oct 08 21:23:01 mesos rexray[21340]: main.main()
Oct 08 21:23:01 mesos rexray[21340]: /home/travis/gopath/src/github.com/emccode/rexray/rexray/rexray.go:87 +0x458

[root@host ~]# ls /var/run/libstorage.sock
/var/run/libstorage.sock
```

akutz:
Closed with #642.

PR #642, opened 2016-12-02
Auto-Detect Running Service

akutz:
This patch handles issue #589 so that a REX-Ray client will automatically detect a running libStorage service if it was started by REX-Ray, even via embedded mode. This means that even if the REX-Ray configuration file relies on auto-service mode where a host is not explicitly defined, if a libStorage server is started in this manner, subsequent REX-Ray executions will not start additional servers, instead attaching to the existing service.

- [x] Successful test result
  - [x] @cduchesne 
  - [x] @akutz

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -5,10 +5,12 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path"
 	"regexp"
 	"strconv"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	gofig "github.com/akutz/gofig/types"
 	"github.com/akutz/gotil"
 	apiversion "github.com/codedellemc/libstorage/api"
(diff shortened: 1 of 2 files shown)


util/util.go
```diff
@@ -167,29 +171,37 @@ func BinDirPath() string {
 // PidFilePath returns the path to the REX-Ray PID file.
 func PidFilePath() string {
 	if pidFilePath == "" {
-		pidFilePath = fmt.Sprintf("%s/rexray.pid", RunDirPath())
+		pidFilePath = RunFilePath("rexray.pid")
 	}
 	return pidFilePath
 }
 
+// SpecFilePath returns the path to the REX-Ray spect file.
```
codenrhoden: minor typo in comment.  "spect" vs "spec"

akutz:
Oh come on. :) thanks for catching it :)

-- 
-a

PR #649, opened 2016-12-06
Support for Creating Encrypted Volumes

akutz:
This patch updates the `volume create` command to support the new flag `--encrypted`. When specified, this flag sends an option to the libStorage volume creation API that requests the creation of a new storage volume using encryption. This feature does not guarantee back-end encryption -- it's up to the libStorage storage driver and underlying storage platform to support encryption.

### DIFF
--- a/cli/cli/cli.go
+++ b/cli/cli/cli.go
@@ -127,6 +127,7 @@ type CLI struct {
 	moduleInstanceAddress   string
 	moduleInstanceStart     bool
 	moduleConfig            []string
+	encrypted               bool
 }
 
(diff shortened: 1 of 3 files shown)


akutz:
This PR was created as a result of this Slack conversation:

image

akutz:
Hi @cduchesne,

If you look at the code you'll see the compromise I made with showing `Encrypted` as one of the default fields. If a returned volume list has any volume with the `Encrypted` field set to true, the `Encrypted` column is included in the output. Otherwise it is omitted.

PR #654, opened 2016-12-07
Fix Volume Status for Detach Op

akutz:
This patch fixes the issue where a volume's status is not accurately reflected upon being detached as well as a few other operations re: the volume status.

### DIFF
--- a/cli/cli/cmds_volume.go
+++ b/cli/cli/cmds_volume.go
@@ -151,6 +151,7 @@ func (c *CLI) initVolumeCmds() {
 					}
 					if c.attach || c.amount {
 						dv.Attachments = withAttachments
+						dv.AttachmentState = apitypes.VolumeAttached
 					}
 					if c.amount {
 						processedPVols = append(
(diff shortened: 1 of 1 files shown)


akutz:
I validated this on an EC2 instance with an encrypted volume.

```
[root@ip-172-31-16-225 rexray]# rexray -l error volume ls
ID            Name    Status     Size  Encrypted
vol-9488b51c          attached   8     false
vol-ec8eb364          available  8     false
vol-1b84b993  builds  available  16    false
vol-d580bd5d  logs    available  1     true
vol-8a7c4002  logs2   available  1     true

[root@ip-172-31-16-225 rexray]# rexray -l error volume attach logs2
ID            Name   Status    Size  Encrypted
vol-8a7c4002  logs2  attached  1     true

[root@ip-172-31-16-225 rexray]# rexray -l error volume ls --attached
ID            Name   Status    Size  Encrypted
vol-9488b51c         attached  8     false
vol-8a7c4002  logs2  attached  1     true

[root@ip-172-31-16-225 rexray]# rexray -l error volume detach logs2 -n
ID            Name   Status     Size  Encrypted
vol-8a7c4002  logs2  available  1     true

[root@ip-172-31-16-225 rexray]# rexray -l error volume detach logs2 
ID            Name   Status     Size  Encrypted
vol-8a7c4002  logs2  available  1     true
```

PR #664, opened 2016-12-12
Classic, Client, Agent, & Controller Builds

akutz:
This patch updates REX-Ray and the build scripts so that REX-Ray now supports four build types: classic, client, agent, & controller.

image

## Build Types
Build Type | File Name | Client | Docker Volume Plug-in Endpoint Service | libStorage Service
----------|------------|:-----:|:--------------------------------:|:--------:
Classic | `rexray` | &#9989; | &#9989; | &#9989;
Client | `rexray-client` | &#9989; | |
Agent | `rexray-agent` | | &#9989; |
Controller | `rexray-controller` | | | &#9989;

## How To Build
The above REX-Ray types can be built with the following commands:

Build Type | Commands
-----------|------------
Classic | `$ make` <br/>_or_<br/> `$ make deps && make build` <br/>_or_<br/> `$make docker-build`
Client | `$ make deps && make build-client` <br/>_or_<br/> `$make docker-build-client`
Agent | `$ make deps && make build-agent` <br/>_or_<br/> `$make docker-build-agent`
Controller | `$ make deps && make build-controller` <br/>_or_<br/> `$make docker-build-controller`

## Cleaning Between Builds
Unless the Docker option is used to build REX-Ray, it is necessary to use `$ make clobber` to create a clean environment before building  another type. This is because build tags are used to control which  build is produced, and the build system is not able to consider build  tags when considering which targets need to be rebuilt. Thus if the  REX-Ray core package is already built in `client` mode and someone  attempts to build REX-Ray for `controller` mode, the core package  will not be rebuilt and may be incorrect for `controller` mode.

## Docker Builds
However, the above issue is not a concern when building REX-Ray  with Docker since each build type uses a Docker container with the prefix `rexray-build` and an eponymous suffix. For example, the  container used to build the REX-Ray client is named  `rexray-build-client` whereas the container used to build the  controller is named `rexray-build-controller`. The classic REX-Ray build is still built with a container named `build-rexray`.

## Build Type Info
There are targets named `info`, `info-client`, `info-agent`, and  `info-controller` that can be used to obtain information about  each of the different build scenarios. There are also Docker  variants of these targets. For example, `docker-info`,  `docker-info-client`, etc.

## Installing
While nothing changes for the classic build type, the new build types have some changes with regards to installation.

### Client Installation
The client build still supports the `install` and `uninstall` commands so that the `rexray-client` binary can be installed/uninstalled via the RPM/DEB package management systems, but the actual commands do nothing. 

### Agent Installation
The `rexray-agent` build is installed just as the classic mode, with the exception that the agent's log and PID files are now named `rexray-agent.log` and `rexray-agent.pid`. All other data directories and behaviors are the same. 

### Controller Installation
The `rexray-controller` build is installed just as the classic mode, with the exception that the controller's log and PID files are now named `rexray-controller.log` and `rexray-controller.pid`. All other data directories and behaviors are the same.


util/util_test.go
```diff
@@ -119,12 +119,12 @@ func TestPrefixAndDirs(t *testing.T) {
 		t.Fatalf("BinDirPath() == %s, != %s", binDirPath, expBinDirPath)
 	}
 
-	binDirFilePath := BinFilePath()
-	expBinFilePath := fmt.Sprintf("%s/%s", binDirPath, "rexray")
-	if expBinFilePath != binDirFilePath {
-		t.Fatalf("BinFilePath(\"rexray\") == %s, != %s",
-			binDirFilePath, expBinFilePath)
-	}
+	//binDirFilePath := BinFilePath()
```
codenrhoden: want to keep these in?

Makefile
```diff
@@ -1,13 +1,61 @@
...
+BUILD_TAGS := $(filter-out libstorage_storage_driver_%,$(BUILD_TAGS))
+BUILD_TAGS := $(filter-out libstorage_storage_executor,$(BUILD_TAGS))
+BUILD_TAGS := $(filter-out libstorage_storage_executor_%,$(BUILD_TAGS))
+endif
+
+ifneq (,$(findstring rexray_build_type_controller,$(BUILD_TAGS)))
+PROG := $(PROG)-controller
+REXRAY_BUILD_TYPE := controller
+BUILD_TAGS := $(filter-out libstorage_integration_driver_%,$(BUILD_TAGS))
+# if this is a controller build then consider the DRIVERS var as it may
```
akutz: This is a bug. Thank you for catching it.

akutz:
Hi @codenrhoden,

Did you begin the review before I updated the PR's description? It should be *far* more clear now. Take a look and let me know. Thanks.

codenrhoden:
@akutz I wish I could say I did. you did push some changes that negated some of the comments I had queued up (getting rid of some commented out code), but the PR description was already in place. Even with the awesome chart that says an agent build only has the docker volume plugin endpoint service, I am too dense to fully understand what this means.

Looking through things, it does seem like a CLI is included with all three new types of builds, and than that a subset of functionality is included based on type. A controller build can't do attach and mount, for example.

I'm just having a hard time picturing the agent build. Naively I would think that it was more client + docker volume plugin. What does the build with only the docker volume plugin end up looking like?

akutz:
Hi @codenrhoden,

The agent-only build exists for a few reasons:

* It enables the client build to be super small since it doesn't include any of the module or docker endpoint code.
* The client build no longer requires root user to run since no socket file must be created for the Docker endpoint service.
* All three builds can be replaced separate from one another. For example -- upgrade the CLI but leave the agent binary alone.

codenrhoden:
Ah, okay, I think I get it.

Should you choose to use the minimal builds (rather than classic), you would ultimately install `rexray-agent` to get the docker volume plugin stuff **AND** `rexray-client` if you wanted the CLI for interfacing with volumes directly through rexray. And as you said, you could update them independently.

Thanks.

akutz:
Hi @codenrhoden,

To be very clear, `rexray-client` doesn't interface with `rexray-agent` in any way as the latter does not start an embedded libStorage server. Eventually if the REX-Ray client adds back functionality to list/modify running modules in the agent then perhaps.

PR #676, opened 2016-12-17
Fix for FlexRex script

akutz:
This patch updates the FlexRex script to no longer use the -q flag for the where command which as it turns out is not POSIX compliant.


akutz:
Related to #641

PR #677, opened 2016-12-17
Makefile Bug & Symlink Error

akutz:
This patch fixes a bug in the Makefile that prevents scripts from being
compiled into the rexray-client binary. Additionally, this patch adds
the --force flag to the "rexray flexrex install" command to enable
creating a symlink to a location that already exists.


akutz:
Related to #641

PR #678, opened 2016-12-18
Fix for bad jq command input

akutz:
This patch fixes an issue where a jq command input was not a properly
formatted JSON query due to single-quotes in lieu of double.


akutz:
Related to #641

Issue #552, opened 2016-09-08
Update REX-Ray documentation to reflect new company name


jonasrosland:
The team name is now "{code} by Dell EMC" instead of "EMC {code}"

PR #686, opened 2017-01-03
Changed Storage Provider Support from EMC to Dell EMC

kacole2:
As the title said, made this change in the documentation in two spots.

Signed-off-by: Kendrick Coleman <kendrickcoleman@gmail.com>


kacole2:
This fixes #552

PR #694, opened 2017-01-12
Fix usage of make in Vagrant file for new build types

codenrhoden:
A simple "make" in this Vagrantfile was working anymore.
Use "make build" instead.

Update to newer version of Golang as well.


codenrhoden:
I didn't personally run into any memory/RAM issues when building, either.

Issue #663, opened 2016-12-12
Documentation for K8s FlexREX integration


cduchesne:
We need to add some documentation for using REX-Ray with the Kubernetes Flexvolume plugin.

#714

cduchesne:
thanks @vladimirvivien. #714 merged

PR #714, opened 2017-02-08
Documentation for FlexRex integration (see #663)

vladimirvivien:
This commit adds documentation for FlexRex integration and includes
examples of persistent volume, persitent volume claim, and pods.

### DIFF
--- a/.docs/user-guide/schedulers.md
+++ b/.docs/user-guide/schedulers.md
@@ -155,6 +155,178 @@ $ docker volume rm vbox2
 Please review the [Applications](./applications.md) section for information on
 configuring popular applications with persistent storage via Docker and REX-Ray.
 
+## Kubernetes
+REX-Ray can be integrated with [Kubernetes](https://kubernetes.io/) allowing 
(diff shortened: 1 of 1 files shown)


clintonskitson:
Would it make sense to include the parameters such as forceAttach?

vladimirvivien:
@clintonskitson the YAML includes `forceAttach` params. Do you mean including somewhere else ?

cduchesne:
I agree that a table describing the options could be helpful.

vladimirvivien:
@clintonskitson @cduchesne doesnt `forceAttach` have a universal meaning in REX-Ray ?

cduchesne:
@vladimirvivien yes but good to connect forceAttach and forceAttach delay to what they mean here as well. In REX-Ray we have a --force flag available to the attach command, although this may not always be clearly visible to some. We definitely have no concept of the forceAttachDelay yet.

vladimirvivien:
@clintonskitson @cduchesne added doc for options, update to the pre-reqs.

vladimirvivien:
@cduchesne PTAL

Issue #719, opened 2017-02-10
Alpine Linux support


lmakarov:
# Summary

Add support for Alpine Linux

# New Feature

Docker for AWS (Docker for Mac and Windows) provisions nodes using Docker's Moby Linux (Alpine Linux). Adding support for Alpine will make it possible to use REX-Ray with the stock Docker VM images.

akutz:
Hi @lmakarov,

Could you please provide additional information, such as:

- How does REX-Ray behave today with Alpine Linux?
- What is the expected behavior?
- Do you have it working with tweaks? If so, which changes have you made?

Thank you.

kacole2:
@lmakarov checkout this blog from yesterday. [Persistent storage in Docker for AWS is not a mystery}(https://blog.codedellemc.com/2017/02/09/persistent-storage-docker-aws-not-a-mystery/)

in addition, here is everything you need for Docker for AWS in the [{code} Labs](https://github.com/codedellemc/labs). 

I have a very hacky way of building Alpine images today but this is NOT what should be used gong forward. Again, this was just a hack to prove it can be done [EBS and EFS Volumes with Docker For AWS using REX-Ray](https://gist.github.com/kacole2/fd425197d62a16c8b09e8ee76660f6b8)

clintonskitson:
It would be great to get some more details as @akutz described.

Another thing is that there is an active conversation in #project-rexray channel in the community where @lax77 is working through the creation of an alpine based container image for REX.

So there are really two things here.

1) Containerizing REX with Alpine as the base custom image
2) Ensuring REX itself is compatible with Alpine as a base OS when not running REX in the container. I believe the containerization (#1) provides the insight as to what packages need to be installed in this case.

lmakarov:
@kacole2 

>checkout this blog from yesterday

Wow, what a great timing, thank you! I wish the blog post and the video showed up in Google search results earlier, as I spent quite some time in the last 2 days trying to figure out how to use REX on Alpine directly or dockerize it (which I thought was abandoned as an idea after this PR was closed https://github.com/codedellemc/rexray/pull/347).

@akutz 

> How does REX-Ray behave today with Alpine Linux?

There is an number of issues. I figured out some of them, but not all. The [gist](EBS and EFS Volumes with Docker For AWS using REX-Ray) from @kacole2 provides some hints.

**Issue 1:** curl complains about SSL certs

```
~ $ curl -sSL https://dl.bintray.com/emccode/rexray/install | sh -
[...]

PR #724, opened 2017-02-12
Alpine-Based Docker Plug-in / S3FS Plug-in

akutz:
This patch switches the Docker plug-in image to be based on Alpine as well as introduces an S3FS plug-in skeleton. This relates to issue #719. 

This issue is also a work-in-progress, or it may also simply be rebased to no longer include the S3FS support as that presents its own challenges with including a working `s3fs` binary. While there [is a project](https://github.com/rjocoleman/docker-alpine-s3fs) that outlines how to build an Alpine-based image with `s3fs`, it does not work. The system complains that the `fuse` module is unable to be loaded or located by `modprobe` and when attempting to mount the bucket with the `s3fs` binary. So far no research has uncovered a solution.

### DIFF
--- a/.docker/plugins/.gitignore
+++ b/.docker/plugins/.gitignore
@@ -1,5 +1,0 @@
-/*/Dockerfile
-/*/rexray.sh
-/*/rexray.yml
-/.travis-*.json
-/*/rexray
--- a/.docker/plugins/Dockerfile
+++ b/.docker/plugins/Dockerfile
@@ -1,11 +1,12 @@
-FROM centos
+FROM alpine:3.5
 
 LABEL drivers="${DRIVERS}"
 LABEL version="${VERSION}"
 
-RUN yum update -y
-RUN yum install xfsprogs e2fsprogs -y
+RUN apk update
+RUN apk add xfsprogs e2fsprogs ca-certificates
 
+RUN mkdir -p /lib64 && ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
 RUN mkdir -p /etc/rexray /run/docker/plugins /var/lib/libstorage/volumes
 ADD rexray /usr/bin/rexray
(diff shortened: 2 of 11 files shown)


akutz:
Hey @cduchesne,

Please test this with EBS to see if it works for you. I had an issue with S3FS until I added the ca-certificates package. After that it worked fine (up until the mount, but that was unrelated), but I was testing inside a running container, not as the plug-in.

cduchesne:
@akutz I was able to get this working great once I added the bash package and removed the tee command from the exec. Let me know when you push these changes and I'll retest just to be sure.

cduchesne:
@akutz I just finished testing changes... works well

thoughts on rexray.yaml: is `service: ${DRIVERS}` valid? Wouldn't this depend on DRIVERS only containing a single driver? Also, what is purpose of the `server: services:` portion?

akutz:
Some files are copied into staging, some are manipulated. Read the Makefile, watch the output when the plugin is built, or see the produced config inside staging area. Regarding DRIVERS, I take your concern into account already. Read the Makefile.

-- 
-a

cduchesne:
@akutz ahh thanks makes sense - i am good with this

akutz:
Hi @cduchesne,

FWIW, the question about the multiple DRIVERS is answered in this PR's diff actually:

image

And the question about the services line ending with no services?

image

cduchesne:
@akutz i checked the diff right after... i was more confused by the wording DRIVERS even though modified by Makefile. It all makes sense, sorry for not reading through each change.

Thanks for the second piece... this is for when we support multiple drivers in a single plugin in the future... got it.

akutz:
Hi @cduchesne,

Yeah, I just persisted the same name throughout to keep the value consistent. I agree it's misleading. Other than that one property actually, multiple drivers actually don't present a problem. they just slot right in. For that I had to get cute.

PR #753, opened 2017-02-23
PID Chaos in Docker Plug-in

akutz:
This patch fixes an issue where a stale PID file used by REX-Ray is never properly removed and locks the program due to the Docker plug-in model of fast-restarting a plug-in if misconfigured combined with the entry point always using PID 1.

Additionally, this patch updates the `Dockerfile`s so that the entrypoint is now defined with default arguments that can be overridden via the CLI for test and development purposes.

### DIFF
--- a/.docker/plugins/Dockerfile
+++ b/.docker/plugins/Dockerfile
@@ -14,4 +14,5 @@ ADD rexray.yml /etc/rexray/rexray.yml
 ADD rexray.sh /rexray.sh
 RUN chmod +x /rexray.sh
 
-ENTRYPOINT [ "/rexray.sh", "rexray", "start", "-f" ]
+CMD [ "rexray", "start", "-f", "--nopid" ]
+ENTRYPOINT [ "/rexray.sh" ]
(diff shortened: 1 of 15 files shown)


akutz:
Hi @codenrhoden,

Would you please approve once you realize why your objection makes a fatal assumption. Hint, where are those drivers' `Dockerfile`? :)

akutz:
Hi @codenrhoden,

To not be a dick about it, I'll just say what is occurring. I know you're busy; too busy for my silly games. There is a default `Dockerfile`. You can override any of the default files per-driver by creating the same file in the driver's directory with a leading `.` prefix. The EBS and GCEPD drivers simply don't override the default `Dockerfile`, which *has* been updated.

akutz:
Hi All,

@cduchesne pointed out that when building a plug-in the `Dockerfile`'s entrypoint is ignored in favor of the `config.json` file's entry point. I've updated those now as well, and changed the entrypoint in the `Dockerfile`s to `["tail", "-f", "/dev/null"]` for testing. This way it's easy to launch a plug-in image as detached container and exec into it to test things during development.

cduchesne:
@akutz tested as-is and works great. Thanks for providing a --nopid option

akutz:
Hi @cduchesne,

Can you please review the latest change to make sure this is what we discussed? Thanks!

PR #758, opened 2017-02-24
Docker Plugin Docs for EFS, S3FS, Isilon, GCEPD

cduchesne:
Adding documentation for EFS, S3FS, Isilon, and GCEPD docker plugins.

### DIFF
--- a/.docs/user-guide/docker-plugins.md
+++ b/.docs/user-guide/docker-plugins.md
@@ -70,7 +70,7 @@ allow-all-devices | `true`
 capabilities | `CAP_SYS_ADMIN`
 
 #### Configuration
-The following environment variables can be used to configure the ScaleIO
+The following environment variables can be used to configure the EBS
 plug-in:
 
 Environment Variable | Description | Default | Required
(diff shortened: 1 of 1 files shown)


.docs/user-guide/docker-plugins.md
```diff
@@ -80,18 +80,118 @@ Environment Variable | Description | Default | Required
...
+---------------------|-------------|---------|---------
+`EFS_ACCESSKEY` | The AWS access key | | ✓
+`EFS_SECRETKEY` | The AWS secret key | | ✓
+`EFS_REGION` | The AWS region | |
+`EFS_SECURITYGROUPS` | The AWS security groups to bind to | `default` |
+`EFS_TAG` | Only consume volumes with tag (tag\volume_name)| |
+`EFS_DISABLESESSIONCACHE` | new AWS connection is established with every API call | `false` |
 
-### Simple Storage Service
+### Simple Storage Service (S3)
```
codenrhoden: Is any of our documention linking here directly? if so, the link to it will break (unless you add of the new hidden anchors).  As this stuff is so new, I don't think there are any links to this stuff yet.

.docs/user-guide/docker-plugins.md
```diff
@@ -147,11 +247,38 @@ Environment Variable | Description | Default | Required
...
 
-### Azure Unmanaged Disk
+#### Configuration
+The following environment variables can be used to configure the GCEPD
+plug-in:
+
+Environment Variable | Description | Default | Required
+---------------------|-------------|---------|---------
+`GCEPD_DEFAULTDISKTYPE` | The default disk type to consume | `pd-ssd` |
+`GCEPD_PROJECTID` | GCE Project ID | `true` |
```
codenrhoden: This var does not exist.

.docs/user-guide/docker-plugins.md
```diff
@@ -147,11 +247,38 @@ Environment Variable | Description | Default | Required
...
+`GCEPD_REGION` | GCE Region | `false` |
```
codenrhoden: This var does not exist.  Only `zone` `disktype` and `tag`.  So just delete this and the one before it.

PR #785, opened 2017-03-24
PlantUML Documentation

akutz:
This patch includes PlantUML (http://plantuml.com/) documentation for common REX-Ray workflows.


codenrhoden:
Hey @akutz.

This is awesome. I like it. I got my Atom plugin installed -- super slick.

Visually, I prefer all the participants that are within the same system/machine/OS to be in the same box. I'm curious what you think. It was an easy change for me to play with, so what do you think about something like this instead:

volume-create

akutz:
Hi @codenrhoden,

The reason the PUML displays the groupings the way it does is because I preferred to organize by process instead of by host since I was unable to do both. If you can find a way to both then I'm all for the change.

codenrhoden:
Yeah, I didn't have any luck either. Their `rectangle` keyword doesn't appear to apply to sequence diagrams, and all of their feature requests for nested boxes have gone unanswered. *sigh*.

akutz:
Hi @codenrhoden,

So you see this is why I elected for division by process. I feel it's more accurate as a lot of the isolation and modularity occurs at a process, and not host, level.

akutz:
Hi @codenrhoden,

Any objections to merging this? Or do you feel like per-process is not appropriate?

codenrhoden:
@akutz No objection. If PUML gains an enhancement to get what we both would like to see, we can revisit! :) Thanks again for putting this together.

PR #786, opened 2017-03-27
Use upstream Cobra package

### LINKED ISSUE #743: Explicitly setting --attached=false returns attached vols
# Summary
As options to `rexray volum ls`, the help shows the following:

```
      --attached[=false]: A flag that indicates only volumes attached to this host should be returned
      --available[=false]: A flag that indicates only available volumes should be returned
      --path[=false]: A flag that indicates only volumes attached to this host should be returned, along with their path info
```

This indicates that `--attached=false` is an option. But when you use it, attached volumes still get returned.

## Version
```shell
$ rexray version
REX-Ray
-------
Binary: /usr/bin/rexray
SemVer: 0.8.0-rc1
OsArch: Linux-x86_64
Branch: v0.8.0-rc1
Commit: 483caefe9a904163daaf82c74a7652254415fce5
Formed: Tue, 21 Feb 2017 20:15:45 UTC

libStorage
----------
SemVer: 0.5.0-rc2
OsArch: Linux-x86_64
Branch: v0.8.0-rc1
Commit: 4a1b372134e27da77bbfa109ad5dac0e9499fbc4
Formed: Tue, 21 Feb 2017 20:14:29 UTC
```

## Expected Behavior
That setting `--attached=false` excludes attached volumes.

## Actual Behavior
attached volumes are returned.

```
root@rexray1:~# rexray volume -l warn ls --path
ID                                    Name  Status    Size  Path
de89b811-f928-11e6-98b3-000f5339d321  rex1  attached  1

root@rexray1:~# rexray volume -l warn ls --attached
ID                                    Name  Status    Size
de89b811-f928-11e6-98b3-000f5339d321  rex1  attached  1

root@rexray1:~# rexray volume -l warn ls --attached=false
ID                                    Name  Status    Size
de89b811-f928-11e6-98b3-000f5339d321  rex1  attached  1

root@rexray1:~# rexray volume -l warn ls --available
ID  Name  Status  Size

root@rexray1:~# rexray volume -l warn ls --available=false
ID                                    Name  Status    Size
de89b811-f928-11e6-98b3-000f5339d321  rex1  attached  1
```

Looking at the code, its clear why this is the case.

```golang
			if !c.volumePath {
				attReqType := apitypes.VolAttReq
				if c.volumeAttached {
					attReqType = apitypes.VolAttReqOnlyVolsAttachedToInstance
				} else if c.volumeAvailable {
					attReqType = apitypes.VolAttReqOnlyUnattachedVols
				}
				c.mustMarshalOutput(c.lsVolumes(args, attReqType))
				return
			}
```

We only request `OnlyUnattachedVols` if the `--available` flag has been passed. Since the default value for `c.VolumeAttached` is already `false`, there is no way to detect if someone has explicity set `--attached=false` at the CLI. Not sure what the solution is here, but it's an inconsistency/unexpected behavior if you read the help output.

codenrhoden:
We had been using a fork of Cobra to allow us to redefine the shorthand flag for hep (using "?" instead of "h"). The upstream package changed the way it auto-added the help flags that allows us to go back to using the upstream package now.

Switching to the upstream packages fixes #743.  The problem there was always that the help output implied that the boolean flag could be used as either true *or* false. In reality, we default to false, and only take action when set to true. Explicitly setting to false is not intended to affect a behavior.

Before this change:

```
rexray help volume ls
....
Flags:
      --attached[=false]: A flag that indicates only volumes attached to this host should be returned
      --available[=false]: A flag that indicates only available volumes should be returned
      --path[=false]: A flag that indicates only volumes attached to this host should be returned, along with their path info
...
```

After:

```
rexray help volume ls
...
Flags:
      --attached    A flag that indicates only volumes attached to this host should be returned
      --available   A flag that indicates only available volumes should be returned
      --path        A flag that indicates only volumes attached to this host should be returned, along with their path info
...
```

### DIFF
--- a/cli/cli/cmds_00_other.go
+++ b/cli/cli/cmds_00_other.go
@@ -69,8 +69,8 @@ func (c *CLI) initOtherCmds() {
 }
 
 func (c *CLI) initOtherFlags() {
-	cobra.HelpFlagShorthand = "?"
-	cobra.HelpFlagUsageFormatString = "Help for %s"
+	c.c.PersistentFlags().BoolP("help", "?", false,
+		"Help about the current command")
 
 	c.c.PersistentFlags().StringVarP(&c.cfgFile, "config", "c", "",
 		"The path to a custom REX-Ray configuration file")
--- a/cli/cli/usage.go
+++ b/cli/cli/usage.go
@@ -161,7 +161,7 @@ Aliases:
 Examples:
 {{.Example | rtrim}}{{end}}{{ if .HasAvailableSubCommands}}
 
-Available Commands: {{range cmds $cmd}}{{if (not .IsHelpCommand)}}
+Available Commands: {{range cmds $cmd}}{{if (not .IsAdditionalHelpTopicCommand)}}
   {{rpad .Name .NamePadding }} {{.Short | rtrim}}{{end}}{{end}}{{end}}{{$lf := lf $cmd}}{{if hf $lf}}
 
 Flags:
@@ -174,7 +174,7 @@ Global Flags:
 {{$fs.FlagUsages | rtrim}}
 {{end}}{{end}}{{if .HasHelpSubCommands}}
 
-Additional help topics: {{range .Commands}}{{if .IsHelpCommand}}
+Additional help topics: {{range .Commands}}{{if .IsAdditionalHelpTopicCommand}}
   {{rpad .CommandPath .CommandPathPadding}} {{.Short | rtrim}}{{end}}}{{end}}{{end}}{{if .HasSubCommands}}
 
 Use "{{.CommandPath}} [command] --help" for more information about a command.{{end}}
--- a/glide-docker.yaml
+++ b/glide-docker.yaml
@@ -24,11 +24,9 @@ import:
     version: v0.1.0
 
   - package: github.com/spf13/pflag
-    ref:     b084184666e02084b8ccb9b704bf0d79c466eb1d
-    repo:    https://github.com/spf13/pflag
+    ref:     d16db1e50e33dff1b6cdf37596cef36742128670
   - package: github.com/spf13/cobra
-    ref:     363816bb13ce1710460c2345017fd35593cbf5ed
-    repo:    https://github.com/akutz/cobra
+    ref:     7be4beda01ec05d0b93d80b3facd2b6f44080d94
   - package: golang.org/x/net
     repo:    https://github.com/golang/net
 
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 35113c02c0fb913e440780ba5e0590ef88ba132a21c3f895beb77869d3cf862b
-updated: 2017-03-28T14:21:50.729273278-05:00
+hash: 2f83cdcffda2c664b856d5d79927eb548124f57574f8ccf3e675338d8a459e1c
+updated: 2017-03-29T11:30:02.898046438-07:00
 imports:
 - name: cloud.google.com/go
   version: e4de3dc4493f142c5833f3185e1182025a61f805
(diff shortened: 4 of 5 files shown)


cli/cli/cmds_00_other.go
```diff
@@ -69,8 +69,7 @@ func (c *CLI) initOtherCmds() {
 }
 
 func (c *CLI) initOtherFlags() {
-	cobra.HelpFlagShorthand = "?"
-	cobra.HelpFlagUsageFormatString = "Help for %s"
+	c.c.PersistentFlags().BoolP("help", "?", false, "help for "+c.c.Name())
```
akutz: Hi @codenrhoden,

So I've been trying to get back to mainline Cobra for some time. One of the reasons for the fork was, as you mention, the Help flag situation. I've been aware now for a little bit that they've updated the code to enable a customizeable help flag shorthand. However, one of the outstanding issues is something it doesn't appear you've circumvented with this PR, and I'm attaching this message at the point where I see the problem.

> `"help for "+c.c.Name()`

What do you think that prints? I'm pretty sure it will emit `help for rexray`, which is not appropriate when the help text is emitted for a sub-command. That's why I introduced the `HelpFlagUsageFormatString` field to enable the replacement of a `%s` with the name of the ***active*** command.

The only way to achieve the same using mainline cobra is to define the Help flag as a local flag for *all* defined commands and set the usage text of each and every one to `help for COMMAND_NAME` where `COMMAND_NAME` is the command for which the flag is being defined.

akutz: Hi @codenrhoden,

So I've been trying to get back to mainline Cobra for some time. One of the reasons for the fork was, as you mention, the Help flag situation. I've been aware now for a little bit that they've updated the code to enable a customizeable help flag shorthand. However, one of the outstanding issues is something it doesn't appear you've circumvented with this PR, and I'm attaching this message at the point where I see the problem.

> `"help for "+c.c.Name()`

What do you think that prints? I'm pretty sure it will emit `help for rexray`, which is not appropriate when the help text is emitted for a sub-command. That's why I introduced the `HelpFlagUsageFormatString` field to enable the replacement of a `%s` with the name of the ***active*** command.

The only way to achieve the same using mainline cobra is to define the Help flag as a local flag for *all* defined commands and set the usage text of each and every one to `help for COMMAND_NAME` where `COMMAND_NAME` is the command for which the flag is being defined.

codenrhoden:
@akutz 

I did some more testing with this, using GCE, and didn't run into any CLI problems. I tested create,remove, attach/detach, mount/unmount, ls, and used flags like `--path`, `--format jsonp`, etc. All printed output as normal.

You and I discussed that the best thing to do here would be to remove the `"help for "+c.c.Name()` part of the help flag initialization, which would just leave the help description blank rather than *always* printing "Help for rexray". Before I do that, I have one other suggestion for you to consider.

If you look at the help descript for the help _command_ itself, here's what it prints:

```# ./rexray
REX-Ray:

Usage:
  rexray [flags]
  rexray [command]

Available Commands:
  adapter     The adapter manager
  env         Print the REX-Ray environment
  flexrex     The FlexVol REX-Ray plug-in manager
```
It is always **Help about any command**. Would you rather see that instead? So to be clear, which is better:

(1)
```
Global Flags:
  -c, --config string     The path to a custom REX-Ray configuration file
  -?, --help
  -h, --host string       The libStorage host.
```

or (2)
```
Global Flags:
  -c, --config string     The path to a custom REX-Ray configuration file
  -?, --help              Help about any command
  -h, --host string       The libStorage host.
```

(2) is consisent with the _help_ command description, but it is repetitive.

codenrhoden:
Hey @akutz. I have this rebased in my local branch. Just wanting to hear from you whether you like the empty help string better, or the string **"Help about any command"**. See my previous comment.

akutz:
I saw. Why don't you change to `Help about the current command` for the flag? Also, please be careful on the rebase due to the dependency conflicts introduced once I merged RR 0.8.2.

codenrhoden:
@akutz I like that idea, re: help text, and have made that change.

As for the dependencies, I hope I've done the right thing. Were you just worried that I would downgrade libStorage? That got caught by the rebase.  But I did do `glide up` and check in a new `glide.lock` to pick up anything new from cobra/pflag. If there is another way I need to approach it, I'll need your guidance.

akutz:
Hi @codenrhoden,

Yeah, I was wanting to make sure you just `rm glide.lock` and performed a `glide up` after you updated / merged `glide.yaml`. The `glide.lock` files cannot merge cleanly since they have a calculated hash at the top based on the contents of the corresponding `glide.yaml` file.

codenrhoden:
@akutz this is ready for your re-review, and has been rebased so that it can be merged.

PR #821, opened 2017-04-27
Support for default TLS behavior in REX-Ray

vladimirvivien:
This commit adds support for detault TLS communication between REX-Ray client and controller:
- Auto creates keypair files of certificate and key
- Interacts with user to confirm host acceptance
- Creates know_hosts files to store host fingerprint

### DIFF
--- a/cli/cli/cli.go
+++ b/cli/cli/cli.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"time"
 
@@ -149,6 +150,9 @@ type CLI struct {
 	encryptionKey           string
 	idempotent              bool
 	scriptPath              string
+	serverCertFile          string
+	serverKeyFile           string
+	serverCAFile            string
 }
 
 const (
@@ -364,6 +368,7 @@ func (c *CLI) preRun(cmd *cobra.Command, args []string) {
 			c.ctx.WithField("cmd", cmd.Name()).Debug(
 				"creating libStorage client")
 			c.r, err = util.NewClient(c.ctx, c.config)
+			err = c.handleKnownHostsError(err)
 		}
 
 		if err != nil {
@@ -450,6 +455,49 @@ func (c *CLI) rrService() string {
 
 func (c *CLI) logLevel() string {
 	return c.config.GetString("rexray.logLevel")
+}
+
+// handles the known_hosts error,
+// if error is ErrKnownHosts, stop execution to prevent unstable state
+func (c *CLI) handleKnownHostsError(err error) error {
+	if err != nil {
+		urlErr, ok := err.(*url.Error)
+		if !ok {
+			return err
+		}
+		hostErr, ok := urlErr.Err.(*apitypes.ErrKnownHost)
+		if !ok {
+			return err
+		}
+
+		pathConfig := context.MustPathConfig(c.ctx)
+		knownHostPath := pathConfig.UserDefaultTLSKnownHosts
+		if !util.AssertTrustedHost(
+			c.ctx,
+			hostErr.PeerHost,
+			hostErr.PeerAlg,
+			hostErr.PeerFingerprint) {
+			fmt.Println("Aborting request, remote host not trusted.\n")
+			os.Exit(1)
+		}
+
+		if addHostErr := util.AddKnownHost(
+			c.ctx,
+			knownHostPath,
+			hostErr.PeerHost,
+			hostErr.PeerAlg,
+			hostErr.PeerFingerprint,
+		); addHostErr == nil {
+			fmt.Printf("Permanently added host %s to known_hosts file %s\n",
+				hostErr.PeerHost, knownHostPath)
+			fmt.Println("It is safe to retry your last rexray command.\n")
+		} else {
+			fmt.Println("Failed to add entry to known_hosts file: ", addHostErr)
+		}
+
+		os.Exit(1) // do not continue
+	}
+	return err
 }
 
 func store() apitypes.Store {
--- a/util/util_tls.go
+++ b/util/util_tls.go
@@ -1,0 +1,203 @@
+package util
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/akutz/gotil"
+	apitypes "github.com/codedellemc/libstorage/api/types"
+)
+
+var (
+	orgName       = "libstorage"
+	certBlockType = "CERTIFICATE"
+	keyBlockType  = "RSA PRIVATE KEY"
+)
+
+// CreateSelfCert creates self-signed certificate and a private key
+func CreateSelfCert(ctx apitypes.Context, certPath, keyPath, host string) error {
+
+	// if files exist, ignore
+	_, cerErr := os.Stat(certPath)
+	_, keyErr := os.Stat(keyPath)
+	if cerErr == nil && keyErr == nil {
+		ctx.WithFields(log.Fields{
+			"host":     host,
+			"certPath": certPath,
+			"certKey":  certPath,
+		}).Debug("skipping self-cert creation, files exist")
+		return nil
+	}
+
+	certRoot := filepath.Dir(certPath)
+	keyRoot := filepath.Dir(keyPath)
+	if err := os.MkdirAll(certRoot, 0755); err != nil {
+		ctx.WithFields(log.Fields{
+			"host":     host,
+			"certRoot": certRoot,
+		}).Debug("created dir")
+
+		return err
+	}
+	if keyRoot != certRoot {
+		if err := os.MkdirAll(keyRoot, 0755); err != nil {
+			ctx.WithFields(log.Fields{
+				"host":    host,
+				"keyRoot": keyRoot,
+			}).Debug("created dir")
+			return err
+		}
+	}
+
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return err
+	}
+
+	tmpl := x509.Certificate{
+		SerialNumber: serialNumber,
+
+		Subject: pkix.Name{
+			Organization: []string{orgName},
+			CommonName:   host,
+		},
+
+		IPAddresses: []net.IP{net.ParseIP(host)},
+		DNSNames:    []string{"localhost"},
+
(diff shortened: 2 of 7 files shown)


cli/cli/cli.go
```diff
@@ -452,6 +457,48 @@ func (c *CLI) logLevel() string {
 	return c.config.GetString("rexray.logLevel")
 }
 
+func (c *CLI) isConnSecured() {
```
akutz: Hi @vladimirvivien,

I just searched for this function against this PR's diff and didn't see it pop up anywhere else. Is this for debugging only? It does not appear to be used.

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+	}).Debug("self-cert files created")
+
+	return nil
+}
+
+func AssertTrustedHost(ctx apitypes.Context, host, algo string, fingerprint []byte) bool {
+	trusted := "no"
+	fmt.Printf("Rejecting connection to unknown host %s.\n", host)
+	fmt.Printf("SHA Fingerprint presented: %s:%x.\n", algo, fingerprint[0:32])
+	fmt.Print("Do you want to save host to knwon_hosts file? (yes/no): ")
```
akutz: Hi @vladimirvivien,

Oops, a small spelling error on the filename `kwon_hosts`.

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+	fmt.Scanf("%s", &trusted)
+	if strings.ToLower(trusted) == "yes" {
```
akutz: Hi @vladimirvivien,

A couple of comments here:

1. Please refrain from ever using `strings.To(Lower|Upper) ==` to compare string values. Here is an [old](https://github.com/codedellemc/rexray/pull/540#r76169894) comment I made to Proud about the topic:

> I recently read [an article](https://medium.com/@benbjohnson/go-walkthrough-bytes-strings-packages-499be9f4b5bd) by @benbjohnson on ways to optimize Go code, and string comparisons are one area that can be greatly improved. To achieve efficient, case-insensitive string comparison, please use [`string.EqualFold`](https://golang.org/pkg/strings/#EqualFold) instead. Memory is allocated with every `strings.ToLower` call, and this is avoided by using the `EqualFold` comparison function. I've started to use it wherever I would normally have used a `ToLower` or `ToUpper` to perform case-insensitive comparison. 

2. Please ensure that variations of `yes` are valid answers as well. My suggestion is to use a regexp like this example from the Golang playground ([run](https://play.golang.org/p/pOGdvRRJyU)). Essentially something like this: ``if ok, _ := regexp.MatchString(`(?i)^y(?:es)?$`, trusted); ok {`` It's case insensitive and accepts just `y|Y` or any cased version of the word `yes`.

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+		return true
+	}
+	return false
+}
+
+func AddKnownHost(ctx apitypes.Context, knownHostPath, host, algo string, fingerprint []byte) error {
+	knownHostPathDir := filepath.Dir(knownHostPath)
+	_, err := os.Stat(knownHostPathDir)
```
akutz: Hi @vladimirvivien,

Please use `github.com/akutz/gotil.FileExists` for this in order to stay consistent with the rest of the project.

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+	khFile, err := os.OpenFile(knownHostPath, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0600)
+	if err != nil {
+		return err
+	}
+	defer khFile.Close()
+
+	fmt.Fprintf(khFile, "%s %s %x\n", host, algo, fingerprint)
+
+	ctx.WithFields(log.Fields{
+		"host": host, "algo": algo, "fingerprint": fmt.Sprintf("%x", fingerprint)},
```
akutz: Hi @vladimirvivien,

Please ensure all code conforms to be soft-wrapped at the 80char mark for readability. The above line can thus be rewritten like this:

```go
	ctx.WithFields(log.Fields{
		"host":		host, 
		"algo":		algo, 
		"fingerprint":	fmt.Sprintf("%x", fingerprint),
	}).Debug("fingerprint added to known_hosts file")
```

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return err
+	}
+
+	certFile, err := os.Create(certPath)
+	if err != nil {
+		return err
+	}
+	defer certFile.Close()
+	if err := pem.Encode(certFile, &pem.Block{Type: certBlockType, Bytes: certBlock}); err != nil {
```
akutz: Hi @vladimirvivien,

Please ensure all code is written with an 80 character boundary and wrapped accordingly with emphasis on readability. Thank you.

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+		return err
+	}
+	defer keyFile.Close()
+
+	keyBlock := x509.MarshalPKCS1PrivateKey(privKey)
+	if err != nil {
+		return err
+	}
+
+	if err := pem.Encode(keyFile, &pem.Block{Type: keyBlockType, Bytes: keyBlock}); err != nil {
```
akutz: Hi @vladimirvivien,

Please ensure all code is written with an 80 character boundary and wrapped accordingly with emphasis on readability. Thank you.

util/util_tls.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+	}
+
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
+	if err != nil {
+		return err
+	}
+
+	tmpl := x509.Certificate{
```
akutz: Hi @vladimirvivien,

Please ensure all code is written with an 80 character boundary and wrapped accordingly with emphasis on readability. Thank you.

cli/cli/cli.go
```diff
@@ -452,6 +457,48 @@ func (c *CLI) logLevel() string {
...
+	if c.r != nil {
+		svr := c.r.API().ServerName()
+		fmt.Printf("*** %s ***\n", svr)
+	}
+}
+
+// handles the known_hosts error,
+// if error is ErrKnownHosts, stop execution to prevent unstable state
+func (c *CLI) handleKnownHostsError(err error) error {
```
akutz: Hi @vladimirvivien,

Please ensure all code is written with an 80 character boundary and wrapped accordingly with emphasis on readability. Thank you.

In this case this function should invert almost all of the `if` blocks to return early instead of the deep indentations currently present.

cli/cli/cli.go
vladimirvivien: Ok I will take a look to see what's is up with this.

util/util_tls.go
vladimirvivien: Ok.

cli/cli/cli.go
vladimirvivien: Ok.

util/util_tls.go
vladimirvivien: ok

util/util_tls.go
vladimirvivien: That's why we need reviews. Thanks.

util/util_tls.go
vladimirvivien: Ok I will use case folding.

util/util_tls.go
vladimirvivien: @akutz that strict 80-char rule, is that a Govet or Golint rule ?  In the age of 4K HD displays, limiting code to only 80 characters seems to force unnecessary wraps.  Specially strict-80.  I think between 80 to 120 works.

vladimirvivien:
@akutz thanks for the review.  I will start on your proposed changes and wait for the additional reviews.

util/util_tls.go
vladimirvivien: Sure I will use file `gotil.FileExists`

cli/cli/cli.go
```diff
@@ -452,6 +457,49 @@ func (c *CLI) logLevel() string {
 	return c.config.GetString("rexray.logLevel")
 }
 
+// handles the known_hosts error,
+// if error is ErrKnownHosts, stop execution to prevent unstable state
+func (c *CLI) handleKnownHostsError(err error) error {
```
akutz: Hi @vladimirvivien,

I know you've updated the code since I made my previous remark, but this function still pushes the 80char boundary and could be easily fixed if you'd invert the if logic. These nested ifs are not necessary. Just return early.

vladimirvivien:
@akutz that strict 80-char rule, is that a Govet or Golint rule ? In the age of 4K HD displays, limiting code to only 80 characters seems to force unnecessary wraps. Specially strict-80. I think between 80 to 120 works.

akutz:
Hi @vladimirvivien,

That is certainly one opinion. I've enforced this rule for REX-Ray and libStorage since I've been managing them, as well as for other projects I manage. 

Besides increasing readability it also enhances stability and maintainability by making people think about how they structure their code. 

Please invert the if blocks or split up the function in order to account for the width requirement.

Thank you.

vladimirvivien:
@akutz I appreciate the feed back.  
But I still think that 80 chars is compressive and forces line wraps where code
would be otherwise OK. I have also worked and maintained code that did not
use the strict 80-char rule that were just as readable and maintainable.  

What I am saying is that 80-char rule probably should be extended to a 
range of `80 to 120` chars.

Think about it 😄

akutz:
Hi @vladimirvivien,

I appreciate your suggestion about increasing the length, but I've also worked on as well as managed countless projects and have thus considered this issue on many occasions. Just as you see no reason the width cannot be expanded, I see no reason why it shouldn't be respected. In this instance the code block in question is only as long as it is because of the nested `if` logic. Just invert that logic, return early, and Bob's your uncle. :)

util/util.go
```diff
@@ -424,6 +424,20 @@ func ActivateLibStorage(
...
+	certPath := config.GetString(apitypes.ConfigTLSCertFile)
+	keyPath := config.GetString(apitypes.ConfigTLSKeyFile)
+	host := config.GetString(apitypes.ConfigHost)
+
+	ctx.WithFields(log.Fields{
+		"host":     host,
+		"certPath": certPath,
+		"keyPath":  keyPath,
+	}).Debug("generating self-cert")
+	if err := CreateSelfCert(ctx, certPath, keyPath, host); err != nil {
```
akutz: Hi @vladimirvivien,

The host from the config file should not be the host used when generating a cert. The config file property isn't in the correct format for a host name used as a TLS certificate's common name (CN) part of the Subject.

My suggestion is to hardcode this to `127.0.0.1` or let user pick the value via a CLI flag. @cduchesne / @codenrhoden, thoughts?

akutz: Hi @vladimirvivien,

Please relocate this code to the installer and do not generate certs unless the RR binary is a standard or controller binary.

akutz:
Replaced by #829.

PR #839, opened 2017-05-03
Known Host Check Fix or No

akutz:
This patch updates the valid response when answering whether a known host is trusted. Either "yes" or "y" are valid, and they are case-insensitive.


vladimirvivien:
@akutz the change looks good.

akutz:
Hi @vladimirvivien,

Then can you approve your Review?

PR #840, opened 2017-05-03
Use default token expiration of 1 year

codenrhoden:
If user does not specify a token expiration, default to 1 year.

### DIFF
--- a/cli/cli/cmds_60_token.go
+++ b/cli/cli/cmds_60_token.go
@@ -26,6 +26,9 @@ const (
 
 	rrTokExpEnv = "REXRAY_TOKEN_EXP"
 	rrTokExpKey = "rexray.token.exp"
+
+	// 8760 hours is 365 days
+	defaultTokenExp = time.Duration(8760 * time.Hour)
 )
 
 var (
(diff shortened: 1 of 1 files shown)


codenrhoden:
With this patch:

```sh
$ rexray token new --key rexray travis | rexray token decode
Subject  IssuedAt                       NotBefore                      Expires
travis   2017-05-03 12:37:17 -0700 MST  2017-05-03 12:37:17 -0700 MST  2018-05-03 12:37:17 -0700 MST
```

akutz:
Eh, can refactor later if we want.

codenrhoden:
With the existing logic in `tokenValidateExp`, It will only go check the env var and config file if it's not set at the command line. If I set the default flag value to by what we want, it would actually always use that and never check the other places.

This appears to be in contrast with the drivers, who register their config properties, and then those automatically handled across CLI, ENV, and config file. Unless I'm missing something.

akutz:
Oh yeah. Duh. I forgot I didn't make it an actual, registered config property. Thanks. Stupid me.

PR #885, opened 2017-06-07
Module-Scoped Host Definitions

### LINKED ISSUE #827: libstorage.host required in configuration even when defined under modules
When testing modules, I noticed that libstorage.host is required at root of config even if also defined under defined modules.

```
# cat /etc/rexray/rexray-agent.yml
rexray:
  modules:
    default-docker:
      type: docker
      desc: default docker module
      host: unix:///run/docker/plugins/rexray1.sock
      libstorage:
        host: tcp://10.50.0.10:7979
        service: scaleio
    rexray2-module:
      type: docker
      desc: second module
      host: unix:///run/docker/plugins/rexray2.sock
      libstorage:
        host: tcp://10.50.0.20:7979
        service: scaleio2
```

```
ERRO[0000] default module(s) failed to initialize        error=libStorage host config missing time=1493682067471
```

akutz:
This patch fixes #827 where a libStorage host may not be defined at the root of a configuration but instead scoped to a module. With this patch if host detection fails (root configuration / running, local host) when initializing the module subsystem a warning is logged and the modules themselves must explicitly define the host to which to connect.

This patch does *not* support starting a new host from a module-scoped host definition. The module-scoped host properties are for overriding the host to which to connect only.

### DIFF
--- a/daemon/module/module.go
+++ b/daemon/module/module.go
@@ -144,9 +144,16 @@ func InitializeDefaultModules(
 	// enable path caching for the modules
 	config.Set(apitypes.ConfigIgVolOpsPathCacheEnabled, true)
 
+	// attempt to activate libStorage. if an ErrHostDetectionFailed
+	// error occurs then just log it as a warning since modules may
+	// define hosts directly
 	ctx, config, errs, err = util.ActivateLibStorage(ctx, config)
 	if err != nil {
-		return nil, err
+		if err.Error() == util.ErrHostDetectionFailed.Error() {
+			ctx.Warn(err)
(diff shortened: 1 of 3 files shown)


akutz:
Hi @codenrhoden,

The `Scope` call *was* in the `ActivateLibstorage` function as it was common to all components. I relocated it to ensure it wasn't dependent upon LS being activated. It's required so that you're able to redefine `rexray.` prefixed properties inside of other configuration sections.

akutz:
@cduchesne has tested and approved the patch.

PR #888, opened 2017-06-08
Added RR config generator to docs

codenrhoden:
Added the REX-Ray Configuration Generator to the Docs so it will
be a quick way to get going without diving too deep into reading.
In addition, added a QuickConfig section to the Configuration page.

Signed-off-by: Kendrick Coleman <kendrickcoleman@gmail.com>

### DIFF
--- a/.docs/index.md
+++ b/.docs/index.md
@@ -92,30 +92,35 @@ as either a SystemD or SystemV service depending upon the OS.
 $ curl -sSL https://dl.bintray.com/emccode/rexray/install | sh
 ```
 
-
 ### Configure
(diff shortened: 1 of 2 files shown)


codenrhoden:
I changed a few things with this, compared to original patch from @kacole2.

- Changed domain from `emccode` to `codedellemc`. 
- Changed transport from https to http (the generator doesn't have a TLS cert, so browsers throw errors about being insecure).
- Changed text to adhere to 80 char line length where possible (links excepted)
- Fixed invalid link to VirtualBox docs (can't put in a carriage-return mid-link)

Kenny, I'd appreciate you reading through and seeing if you are happy with it still.

PR #893, opened 2017-06-15
Fix Required Permissions for S3FS Docker Plug-In

cduchesne:
This patch fixes the required plug-in permissions for the S3FS Docker Plug-in


cduchesne:
Thanks, updated to add mounts section.

PR #894, opened 2017-06-15
CLI Freeform Options

akutz:
This patch adds support for the `-o|--options` flag to add freeform options to outgoing API requests.


akutz:
See Slack's community channel from the other day

-- 
-a

PR #898, opened 2017-06-19
Introduce Docker managed plugin for Ceph RBD

### LINKED ISSUE #818: Create plugin for Ceph RBD
# Summary
Create a Docker managed volume plugin for Red Hat Ceph from existing `rbd` storage driver.

codenrhoden:
Plugin image requires that we have the Ceph CLI/tools in the container,
so this plugin is built off of CentOS 7 instead of Alpine. As such, the
plugin has customized versions of rexray.sh and Dockerfile instead of
the common ones.

Closes #818.

### DIFF
--- a/.docker/plugins/rbd/.Dockerfile
+++ b/.docker/plugins/rbd/.Dockerfile
@@ -1,0 +1,21 @@
+FROM centos:7.3.1611
+
+LABEL drivers="${DRIVERS}"
+LABEL version="${VERSION}"
+
+ENV CEPH_VERSION kraken
+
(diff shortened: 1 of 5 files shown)


codenrhoden:
This PR requires https://github.com/codedellemc/libstorage/pull/576.

I've tested on CentOS7 and Ubuntu 16.04. Also tested to make sure that this image, which is based on Ceph kraken (latest release), works with a host and cluster install of Ceph Jewel (latest *stable* release). That is likely to be a common scenario, if not the majority of installs.

Issue #776, opened 2017-03-20
CLI Error Improvements


codenrhoden:
As a rexray user, I want to read clear error messages on the CLI that indicate a truer source of problems

As a rexray user, I want to be directed to the libStorage server logs when necessary for further debugging, with a timestamp for better correlation.

Examples of issues today:
- Operations on volumes that exist on the storage platfrom that fail (for various reasons) do not result in a “volume not found”
- Failures resulting from failed CLI executions (e.g. `os.exec()`) do not show the byte reputation of failed commands stdout
- A clear, consistent approach is taken to handling errors generated at the storage driver level
- When to log, what to log
- When to pass error up, when to wrap error
- Wrapped/Inner errors handled fully at rexray client

Questions:
- Need for translation package at rexray layer? This could be contextual -- e.g. what does a 404 usually mean when we do volume attach? It means something different than when we do volume ls
- Are we properly unrolling and summarizing all nested errors returned by libStorage?

Anything that can be broken out into a separate task should be a separate issue, which is why this is created as a tracking Epic.

PR #899, opened 2017-06-20
Log libStorage error from logVolumeLoopError

codenrhoden:
When libStorage would return an error like:
```json
{
  "message": "POST https://api.digitalocean.com/v2/volumes: 422 Storage is not supported for this region",
  "status": 500,
  "error": "POST https://api.digitalocean.com/v2/volumes: 422 Storage is not supported for this region"
}
```

The only thing printed to the REX-Ray CLI output was
"error.status=500", and not the actual error message. Fix that by
asserting type to goof.HTTPError, then extracting the real message.

Fixes codedellemc/libstorage#429

This ends up being the difference between a final output of:
```bash
FATA[0002] error creating volume                         error.status=500 volume=test
```
vs
```bash
FATA[0002] error creating volume                         error.status=500 error.msg=POST https://api.digitalocean.com/v2/volumes: 422 Storage is not supported for this region volume=test
```

I think the new output is much more helpful

### DIFF
--- a/cli/cli/cmds_10_volume.go
+++ b/cli/cli/cmds_10_volume.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/akutz/goof"
 	"github.com/spf13/cobra"
 
 	apitypes "github.com/codedellemc/libstorage/api/types"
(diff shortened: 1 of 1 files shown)


codenrhoden:
I admit I don't fully understand the interaction between golf, goof, and logrus. I tried for about 90 minutes to figure out why something like `log.WithError(httpErr)`, whereby `httpErr` was created with `httpErr, ok := err.(goof.HTTPErr)` wouldn't automatically add `error.msg` to the log fields. `error.status` shows up on its own, but no other fields do.

I tried every combination of `IncludeFieldsWith{Error,String,Format}`, but still no luck. I could log the fields directly that way, by doing `log.Debugf("%+v", httpErr)`, but `log.WithError()` would never show them.

I still think this is a vast improvement, and actually closes out the Epic #776.

PR #917, opened 2017-07-05
WiP - Container Storage Interface (CSI) Module

akutz:
This patch introduces a CSI module that can be configured similarly to the existing Docker module. The CSI module has no default module configuration as there is no default CSI socket file path.

This is a WiP and documentation is forthcoming.


akutz:
Replaced by #975.

PR #934, opened 2017-07-27
Add username param for rbd driver

mvollman:
Add userName parameter to the rbd driver to allow for isolating the
driver to an unpriviledged user with write access to a single rbd
pool.

### DIFF
--- a/libstorage/drivers/storage/rbd/executor/rbd_executor.go
+++ b/libstorage/drivers/storage/rbd/executor/rbd_executor.go
@@ -41,6 +41,10 @@ func (d *driver) Name() string {
 	return rbd.Name
 }
 
+func (d *driver) userName() string {
+	return d.config.GetString(rbd.ConfigUserName)
+}
+
 func (d *driver) Supported(
 	ctx types.Context,
 	opts types.Store) (bool, error) {
@@ -80,7 +84,9 @@ func (d *driver) LocalDevices(
 	ctx types.Context,
 	opts *types.LocalDevicesOpts) (*types.LocalDevices, error) {
 
-	devMap, err := utils.GetMappedRBDs(ctx)
+	username := d.userName()
+
+	devMap, err := utils.GetMappedRBDs(ctx, username)
 	if err != nil {
 		return nil, err
 	}
--- a/libstorage/drivers/storage/rbd/rbd.go
+++ b/libstorage/drivers/storage/rbd/rbd.go
@@ -12,6 +12,9 @@ const (
 	// ConfigDefaultPool is the config key for default pool
 	ConfigDefaultPool = Name + ".defaultPool"
 
+	// ConfigUserName is the config key for rbd username
+	ConfigUserName = Name + ".userName"
+
 	// ConfigTestModule is the config key for testing kernel module presence
 	ConfigTestModule = Name + ".testModule"
 )
@@ -23,6 +26,7 @@ func init() {
 func registerConfig() {
 	r := gofigCore.NewRegistration("RBD")
 	r.Key(gofig.String, "", "rbd", "", ConfigDefaultPool)
+	r.Key(gofig.String, "", "admin", "", ConfigUserName)
 	r.Key(gofig.Bool, "", true, "", ConfigTestModule)
 	gofigCore.Register(r)
 }
--- a/libstorage/drivers/storage/rbd/storage/rbd_storage.go
+++ b/libstorage/drivers/storage/rbd/storage/rbd_storage.go
@@ -72,16 +72,25 @@ func (d *driver) Volumes(
 	ctx types.Context,
 	opts *types.VolumesOpts) ([]*types.Volume, error) {
 
+	username := d.userName()
+
 	// Get all Volumes in all pools
-	pools, err := utils.GetRadosPools(ctx)
+	pools, err := utils.GetRadosPools(ctx, username)
 	if err != nil {
 		return nil, err
+	}
+
+	if username != "admin" {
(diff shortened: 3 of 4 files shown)


akutz:
Hi @codenrhoden,

I assigned you to review this PR as it's related to RBD.

codenrhoden:
Apologies I had not gotten to this yet. I will give it a close look by the end of the week

codenrhoden:
@mvollman Thank you again for taking the time to put this patch together, and again I apologize for being so slow to review it.

We have long known that we needed to add extra config options for accepting a non-default user (admin isn't really a good idea if you've setup your cluster correctly), and there are many other options we might want to take as well.

To this end, I think it would be better to have our command set `CEPH_ARGS`  in the environment when it shells out to the `rados` or `rbd` commands. This would keep us from having to add a new function parameter to many functions for each option we add. This has been requested in https://github.com/codedellemc/rexray/issues/924.

It was also discussed a bit in #957. In there, the OP thought exporting `CEPH_ARGS` to REX-Ray was not ideal, and instead wanted to see a free-form text option in the config file. That would look something like:

```
rbd:
  defaultPool: mypool
  cephArgs: --id myuser
```

I think this might be the best way to go, but it begs the question of what to do if `CEPH_ARGS` is already in the environment and a user has *also* set `cephArgs` in the config file. In my opinion the config file takes precedence, but I could see someone wanting `CEPH_ARGS` to have precedence, or for an error to be thrown. Since the config file is free form text, I would have no intention of trying to merge the two. It may not be a likely scenario, anyways.

So, to recap, I'm suggesting we manually set `cmd.Environment` with what would normally be set in `CEPH_ARGS` from the config option `rbd.cephArgs`. This would be set in places like https://github.com/codedellemc/rexray/blob/1a556e8d457db7166f9446e9d058b29469cd6775/libstorage/drivers/storage/rbd/utils/utils.go#L61

I'd love to hear your thoughts, and whether that is something you would want to try out. If not, please let me know, as I should be able to get that in place with a couple hours of effort, once testing is accounted for. One thing from your patch that I am not sure if it is necessary is limiting the pool to *only* the default pool if not using the `admin` user. A less privileged user could still have access to multiple pools, and I would want them all to be available.  I don't remember what the output of `rados lspools` is if you use a user that only has access to a subset of the pools defined.

codenrhoden:
I'll tackle this in a follow-up release, as it did not make the cut for 0.10.0

mvollman:
My apologies for not getting back to this sooner.  I have been busy and have not been able to prioritize getting back to this since I saw your reply.  I was just trying to get the minimum functionality I needed into the code while maintaining backwards compatibility.  I agree that your suggestion is the better long term approach.  As for allowing a user to see volumes in all pools they have access to, I think that is fine, but `rados lspools` does return all pools not just the pools the user has access to.  The information would likely need to be parsed from ceph auth, assuming the user has permissions to query their capabilities.

codenrhoden:
@mvollman If you have the time, can you check out #1035?

mvollman:
Looks fine to me.  Obviously still a lot of work to do, but 'cephArgs' is certainly a better option then just passing a username.  I did notice that the logic around if cephArgs is set will restrict someone using the admin key to only the default pool if they set a custom cluster name or some other ceph argument in that config option.  Was that intentional?

As I was looking at this again I started digging into possible options for allowing a non-admin user to see all pools that the user has access to.  No good news to report there since ceph appears to restrict all auth queries to the admin user only.  If you wanted to keep that functionality you could go the static config route and along with 'defaultPool' add a list option of 'cephPools' to allow the user to list all the pools they expect to have access to.  Looks like the only dynamic option would be to take a 'ceph osd' or 'rbd' command that would only succeed if the user has access to the pool and iterate over all pools to determine which pools are available to the user.  But that method potentially falls over if a user has read access, but not write access to a given pool.

Let me know how I can help move this forward.

codenrhoden:
@mvollman Thanks for the feedback!

> I did notice that the logic around if cephArgs is set will restrict someone using the admin key to only the default pool if they set a custom cluster name or some other ceph argument in that config option. Was that intentional?

Kind of... As you noted, if you are using a non-admin user, you don't have an easy way of knowing which pools the user has access to. So the intent was to enforce that if you are using a non-admin user, you can only use the one pool. But since CEPH_ARGS is free form, I don't really know what all is set in it, and I used it being not empty to indicate that we should lock it down. But that is extreme, as just settting `--cluster` is something people will want to do. I could instead look at the string for the presence of `--id` or `--name` and only force the defaultPool if those are present. Seem like a good compromise?

> Looks like the only dynamic option would be to take a 'ceph osd' or 'rbd' command that would only succeed if the user has access to the pool and iterate over all pools to determine which pools are available to the user. But that method potentially falls over if a user has read access, but not write access to a given pool.

Yeah, I tried to play with this too. I found that it was really easy, actually, to set a users caps in such a way that simple commands like `rados lspools` hung and never returned, which would lock our entire driver. :( For example, I tried using `profile rbd` for my caps, and that actually didn't work (I felt like I used to). I instead had to use `rwx`. This is likely a bug in whatever version I happened to be using, but it didn't give me confidence in using non-admin users to query pools. So, for now I'd like to stick with the idea that if you use a non-admin user, you can only use the one pool (with the fix from above, which I need to make). Your suggestion of a static list is a good one too, though.

akutz:
Hi @codenrhoden,

I'm unclear after reading the above discussion about the status of this PR? Could you let me know if we should merge this? If so, please go ahead and merge it :)

codenrhoden:
No, this will get replaced by #1035. Sorry for confusion, will close for now.

PR #935, opened 2017-07-27
Fix Cinder support

### LINKED ISSUE #915: Cinder: Mounting volumes in containers doesn't work when the volume isn't attached to the VM
# Summary

I'm trying to use the `rexray/cinder` docker plugin to mount cinder volumes inside containers. So far I've been unable to get this to work unless the cinder volume has been mounted on the VM.

My test docker command is:

```
docker run --rm -it --volume-driver=rexray/cinder  -v test-vol-1:/data busybox sh
```

I've already created by volume using OpenStack.

Running the above command gives the following error:

```
docker: Error response from daemon: error while mounting volume '/var/lib/docker/plugins/98f0aacacd60baa40de8e5aa3fc51cf29ec8d0fae7c5e01812b1e00fcb34e041/rootfs': VolumeDriver.Mount: {"Error":"problem with device discovery"}.
```

However, this command does successfully mount the cinder volume on the VM, just not inside the container. And then running the above command a *second* time works.

If I manually attach the cinder volume to the VM, then the above command works first time. It only seems to fail when the cinder volume *isn't* attached to the VM.

## Version

```
$ docker plugin ls
ID                  NAME                   DESCRIPTION                           ENABLED
98f0aacacd60        rexray/cinder:latest   REX-Ray for Cinder Persistent Disks   true
```

I believe `latest` points to 0.9.2 at the moment.

```
$ docker plugin inspect rexray/cinder
[
    {
        "Config": {
            "Args": {
                "Description": "",
                "Name": "",
                "Settable": null,
                "Value": null
            },
            "Description": "REX-Ray for Cinder Persistent Disks",
            "Documentation": "https://github.com/codedellemc/rexray/.docker/plugin/cinder",
            "Entrypoint": [
                "/rexray.sh",
                "rexray",
                "start",
                "-f",
                "--nopid"
            ],
            "Env": [
                {
                    "Description": "",
                    "Name": "REXRAY_FSTYPE",
                    "Settable": [
                        "value"
                    ],
                    "Value": "ext4"
                },
                {
                    "Description": "",
                    "Name": "REXRAY_LOGLEVEL",
                    "Settable": [
                        "value"
                    ],
                    "Value": "warn"
                },
                {
                    "Description": "",
                    "Name": "REXRAY_PREEMPT",
                    "Settable": [
                        "value"
                    ],
                    "Value": "false"
                },
                {
                    "Description": "",
                    "Name": "CINDER_AUTHURL",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_USERNAMEID",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_USERNAME",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_PASSWORD",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_TOKENID",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_TENANTID",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_TENANTNAME",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_DOMAINID",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                },
                {
                    "Description": "",
                    "Name": "CINDER_DOMAINNAME",
                    "Settable": [
                        "value"
                    ],
                    "Value": ""
                }
            ],
            "Interface": {
                "Socket": "rexray.sock",
                "Types": [
                    "docker.volumedriver/1.0"
                ]
            },
            "Linux": {
                "AllowAllDevices": true,
                "Capabilities": [
                    "CAP_SYS_ADMIN"
                ],
                "Devices": null
            },
            "Mounts": [
                {
                    "Description": "",
                    "Destination": "/dev",
                    "Name": "",
                    "Options": [
                        "rbind"
                    ],
                    "Settable": null,
                    "Source": "/dev",
                    "Type": "bind"
                }
            ],
            "Network": {
                "Type": "host"
            },
            "PropagatedMount": "/var/lib/libstorage/volumes",
            "User": {},
            "WorkDir": "",
            "rootfs": {
                "diff_ids": [
                    "sha256:f6698d039538701f50279c34b67f1efa8c4530c0ca9ddbbf65b39858202f3cee"
                ],
                "type": "layers"
            }
        },
        "Enabled": true,
        "Id": "98f0aacacd60baa40de8e5aa3fc51cf29ec8d0fae7c5e01812b1e00fcb34e041",
        "Name": "rexray/cinder:latest",
        "PluginReference": "rexray/cinder:latest",
        "Settings": {
            "Args": [],
            "Devices": [],
            "Env": [
                "REXRAY_FSTYPE=ext4",
                "REXRAY_LOGLEVEL=debug",
                "REXRAY_PREEMPT=true",
                "CINDER_AUTHURL=https://keystone.****/v3",
                "CINDER_USERNAMEID=",
                "CINDER_USERNAME=****",
                "CINDER_PASSWORD=****",
                "CINDER_TOKENID=",
                "CINDER_TENANTID=****",
                "CINDER_TENANTNAME=",
                "CINDER_DOMAINID=",
                "CINDER_DOMAINNAME=****"
            ],
            "Mounts": [
                {
                    "Description": "",
                    "Destination": "/dev",
                    "Name": "",
                    "Options": [
                        "rbind"
                    ],
                    "Settable": null,
                    "Source": "/dev",
                    "Type": "bind"
                }
            ]
        }
    }
]
```

```
$ docker --version
Docker version 17.03.1-ce, build c6d412e
```

I"m running Docker on Ubuntu 16.04.

## Expected Behavior

The expected behavior is that the cinder volume is mounted on the container even if the volume isn't mounted on the VM before the `docker run` command is run.

## Actual Behavior

The `docker run` command above only works if it is run twice(first time to mount the cinder volume on the VM which exits with an error, second time to actually run the container). Or the `docker run` command will work first time if the cinder volume is already mounted on the VM using OpenStack's tools.

## Steps To Reproduce

1. Create an Ubuntu 16.04 VM in OpenStack
2. Install Docker using the instuctions at https://docs.docker.com/engine/installation/linux/docker-ce/ubuntu/#install-from-a-package
3. Install Rexray:
```
docker plugin install rexray/cinder \
CINDER_AUTHURL=https://keystone.****/v3 \
CINDER_USERNAME=**** \
CINDER_PASSWORD=**** \
CINDER_TENANTID=**** \
CINDER_DOMAINNAME=**** \
REXRAY_PREEMPT=true
```
4. `docker volume create -d rexray/cinder test-vol-1` - this creates a cinder volume but *doesn't* attach it to the current VM
5. `docker run --rm -it --volume-driver=rexray/cinder  -v test-vol-1:/data busybox sh`
6. Shows the above error about device discovery
7. Check OpenStack to see that the newly created volume is now attached to the VM
8. `docker run --rm -it --volume-driver=rexray/cinder  -v test-vol-1:/data busybox sh` now succeeds

## Logs

Since I'm using the docker plugin rather than the Rexray service the logs came from `journalctl`:

https://gist.github.com/agxs/1744c52d45e7e5ee57493d8f39484f5f

MatMaul:
Fix #915

I can't test on master right now, but tested before libstorage merge.
cf codedellemc/libstorage#594

### DIFF
--- a/libstorage/api/types/types_localdevices.go
+++ b/libstorage/api/types/types_localdevices.go
@@ -44,7 +44,9 @@ func (l *LocalDevices) MarshalText() ([]byte, error) {
 	keys := []string{}
 
 	for k := range l.DeviceMap {
-		keys = append(keys, k)
+		if k != "" {
+			keys = append(keys, k)
+		}
 	}
 
 	sort.Sort(byString(keys))
@@ -61,7 +63,7 @@ func (l *LocalDevices) MarshalText() ([]byte, error) {
 }
 
 var (
-	ldRX         = regexp.MustCompile(`^(.+?)=(\S+::\S+(?::\s*,\s*\S+::\S+)*)?$`)
+	ldRX         = regexp.MustCompile(`^(.+?)=(\S+::\S*(?::\s*,\s*\S+::\S*)*)?$`)
 	commaByteSep = []byte{','}
 	colonByteSep = []byte{':', ':'}
 )
@@ -82,10 +84,14 @@ func (l *LocalDevices) UnmarshalText(value []byte) error {
 
 	for _, p := range bytes.Split(m[2], commaByteSep) {
 		pp := bytes.Split(p, colonByteSep)
-		if len(pp) < 2 {
+		if len(pp) < 1 {
 			continue
 		}
-		l.DeviceMap[string(pp[0])] = string(pp[1])
+		val := ""
+		if len(pp) > 1 {
+			val = string(pp[1])
+		}
+		l.DeviceMap[string(pp[0])] = val
 	}
 
 	return nil
--- a/libstorage/api/tests/tests.go
+++ b/libstorage/api/tests/tests.go
@@ -68,6 +68,9 @@ func (sr *suiteRunner) Describe() {
 					t.beforeEachClientServiceSpec()
 				})
 
+				It("preserves local devices with empty values", func() {
+					t.itPreservesLocalDevicesEmptyVals()
+				})
 				It("list volumes", func() {
 					t.itClientSvcSpecListVolumes()
 				})
--- a/libstorage/api/tests/tests_client.go
+++ b/libstorage/api/tests/tests_client.go
@@ -43,3 +43,19 @@ func (t *testRunner) itClientSvcSpecCreateVolume() {
 	t.Θ(t.client.Storage().VolumeCreate(t.ctx, t.volName, t.volCreateOpts()))
 	t.Ε(t.client.Storage().VolumeRemove(t.ctx, t.volID, t.volRemoveOpts()))
 }
+
+func (t *testRunner) itPreservesLocalDevicesEmptyVals() {
+	ld, err := t.client.Executor().LocalDevices(t.ctx, t.locDevOpts())
+	Ω(err).ShouldNot(HaveOccurred())
+	Ω(ld).ShouldNot(BeNil())
+	Ω(ld.DeviceMap).ShouldNot(BeNil())
+	Ω(ld.DeviceMap).ShouldNot(HaveLen(0))
+	preserved := false
+	for _, v := range ld.DeviceMap {
+		if v == "" {
+			preserved = true
+			break
+		}
+	}
+	Ω(preserved).Should(BeTrue())
+}
--- a/libstorage/api/tests/tests_utils.go
+++ b/libstorage/api/tests/tests_utils.go
@@ -76,6 +76,10 @@ func (t *testRunner) volDetOpts() *types.VolumeDetachOpts {
 	return &types.VolumeDetachOpts{Opts: t.store}
 }
 
+func (t *testRunner) locDevOpts() *types.LocalDevicesOpts {
+	return &types.LocalDevicesOpts{Opts: t.store}
+}
+
 func (t *testRunner) Α(vol *types.Volume, tok string, err error) {
 	t.Θ(vol, err)
 	Ω(tok).Should(Equal(t.expectedNextDev))
--- a/libstorage/drivers/storage/cinder/executor/cinder_executor.go
+++ b/libstorage/drivers/storage/cinder/executor/cinder_executor.go
@@ -198,10 +198,28 @@ func (d *driver) NextDevice(
 func (d *driver) LocalDevices(
 	ctx types.Context,
 	opts *types.LocalDevicesOpts) (*types.LocalDevices, error) {
+	devicesMap := make(map[string]string)
 
-	/* There is no current method to look up what the volume ID is for
-	volumes attached to a host via Cinder. All lookups are done on the
-	server-side from the Cinder API
-	*/
-	return &types.LocalDevices{Driver: d.Name()}, nil
+	file := "/proc/partitions"
+	contentBytes, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil,
+			goof.WithFieldE("file", file, "error reading file", err)
+	}
+
+	content := string(contentBytes)
+
+	lines := strings.Split(content, "\n")
+	for _, line := range lines[2:] {
+		fields := strings.Fields(line)
+		if len(fields) >= 4 {
+			devicePath := "/dev/" + fields[3]
+			devicesMap[devicePath] = ""
+		}
+	}
+
+	return &types.LocalDevices{
+		Driver:    cinder.Name,
+		DeviceMap: devicesMap,
+	}, nil
 }
--- a/libstorage/drivers/storage/libstorage/libstorage_client_xcli.go
+++ b/libstorage/drivers/storage/libstorage/libstorage_client_xcli.go
@@ -231,13 +231,6 @@ func (c *client) LocalDevices(
 		return nil, err
 	}
 
-	// remove any local devices without values in the map
-	for k, v := range ld.DeviceMap {
-		if v == "" {
-			delete(ld.DeviceMap, k)
-		}
-	}
-
 	ctx.Debug("xli localdevices success")
 	return ld, nil
 }
--- a/libstorage/drivers/storage/libstorage/libstorage_client_xcli.go
+++ b/libstorage/drivers/storage/libstorage/libstorage_client_xcli.go
@@ -464,12 +464,5 @@ func (c *client) getLocalDevices(
 		return nil, err
 	}
 
-	// remove any local devices without values in the map
-	for k, v := range ld.DeviceMap {
-		if v == "" {
-			delete(ld.DeviceMap, k)
-		}
-	}
-
 	return ld, nil
 }


libstorage/api/types/types_localdevices.go
```diff
@@ -44,7 +44,9 @@ func (l *LocalDevices) MarshalText() ([]byte, error) {
 	keys := []string{}
 
 	for k := range l.DeviceMap {
-		keys = append(keys, k)
+		if k != "" {
```
akutz: Hi @MatMaul,

I would swear on my life I removed the above addition recently. I remember it caused issues when removing the executor as the Marshal function removed the entries sans value. I decided it wasn't the place of a marshal function to do that. But I can find no record I actually made any changes. To be sure this file hasn't been modified in over a year!

The closest I can come is [this](https://github.com/codedellemc/libstorage/commit/2ef3d7edafe33c549fab9b8b016b5e0af66e27c0#diff-a1c38504f420f4720de9cf65512711ea):

image

That's in `drivers/storage/libstorage/libstorage_client_xcli.go`. It removes the devices without values in the map upon receiving the values back from the executor. Would this work for you?

akutz:
Hi @MatMaul,

Did you get a chance to review my comments?

libstorage/api/types/types_localdevices.go
MatMaul: you mean... that :D ?
https://github.com/codedellemc/rexray/pull/935/commits/c1fb815490222077ed94a781da6b5860acda6e84#diff-e8770e421a1e76576af604b2c6441ad2
Either you didn't see the end of my patch or I misunderstood what you said :)
The line you are pointing there avoid empty keys but does not allow empty values, I don't remember why I put that it is probably useless.

akutz:
Hi @MatMaul,

Ah. It looks like mea culpa. You're dealing with the keys and I with the values. I just saw what I was attuned to see given my recent history with the same data. So if we're removing zero-length keys why update the regex to accept them when unmarshalling the same data?

> `ldRX         = regexp.MustCompile(`^(.+?)=(\S+::\S*(?::\s*,\s*\S+::\S*)*)?$`)`

Wouldn't the above pattern prevent a zero-length key or value from even being parsed? Why alter the `+` to `*` so that empty values are even allowed? I get that we have to strip data on both sides, but why allow it to be unmarshalled in the first place?

MatMaul:
The goal is to allow empty values and I think my changes in the regexp do that (I am bad at it so I was perhaps too broad however). You can remove mentally the first chunk of the commit dealing with empty keys and it should make sense.

MatMaul:
I want {"/dev/sda1": ""} possible but not {"": "uuid"}. The second part is functionally useless for my cinder use case (I think).

akutz:
Hi @MatMaul,

That's what I thought. Yes, your pattern is too aggressive, but that's tangential. The issue is that before the executors were removed, local devices with empty values [*were* stripped](https://github.com/codedellemc/libstorage/blob/7368564cc71be9a0dbe03ddfbaeda281232eee72/drivers/storage/libstorage/libstorage_client_xcli.go#L415-L422). When I replaced the executor CLI with code that called the executor implementations directly, the code to strip empty values was removed. This caused the framework tests via the VFS driver to fail. That's why I added the code to remove the local devices with empty values. Except it seems it's actually added in *two* locations: the one you removed and the one in `getLocalDevices`. 

Let's simplify this. We both agree on the following:

* A local device pair with an empty key should *always* be invalid.
* You want to allow empty values
* I'm saying that empty values being stripped cause the VFS tests to fail
* You removed code that strips empty values from `LocalDevices` in the XCLI source but appear to have missed the redundant code in the same file's `getLocalDevices` function.

Ultimately you want pairs with empty values, but even before the executor was removed that case wasn't allowed and there are tests that fail if empty values are allowed.

I guess my question is why the change *now*?

MatMaul:
I completely missed the second method in the same file, sorry for the
confusion.
We are on track now 😊
I am too tired to check now but my guess is that something changed in the
logic of a method calling LocalDevices when checking the attached devices.

akutz:
Hi @MatMaul,

I've rebased your PR onto `master` at https://github.com/akutz/rexray/tree/bugfix/fix-cinder. If you would care to reset this PR to my branch I will merge this PR. I also removed the second location where the local devices with empty values were being removed.

akutz:
Hi @MatMaul,

I also misread your RX changes. It's just fine. I am an idiot across multiple days apparently.

akutz:
Hi @MatMaul,

I also added a test that validates the local devices with empty values are preserved. I'm still concerned, however, that these values *were* removed prior to the executor refactor, and thus any problem that exists now should have existed then as well.

MatMaul:
Done. Yes I guess something changed in the logic validating the device path I'll see if I can pinpoint where it breaks.
Can you have a look at why I can't use rexray with the same CLI/conf as before since the libstorage merge ?
cf https://github.com/codedellemc/libstorage/pull/594
I am unable to test currently.

akutz:
Hi @MatMaul,

I finally figured out the config issue. It was resolved about a week ago, and I didn't connect it to your issue until now while triaging outstanding PRs. The issue was REX-Ray was no longer properly registering the gofig registrations and picking up defaults and registered config properties. That's fixed now if you want to rebase this off of master and give it a whirl.

We're going to try and cut RC1 today, but I've no issue including this in a later RC before release next Monday as you've shown in the past you're quite thorough and produce outstanding code!

MatMaul:
Rebased and tested :+1:

akutz:
Hi @MatMaul,

Great, thanks! Just out of curiosity, is there any reason you cannot squash your commits on this PR? I know you've wanted to keep them separate in the past for purposes of authorship, but is that applicable here?

MatMaul:
2 are yours and I kept 2 on my side because one is a straight revert. But I don't really care I can squash mine together, yours together, everything, let me know.

akutz:
Hi @cduchesne,

I don't think this will make RC1 today, but I'd like to merge it in time for RC2. The reason for my hesitance is because it occurs to me we've only tested against Cinder and VFS. I would like you to test against EBS and ScaleIO if possible to make sure the changes to the local device logic doesn't impact the other drivers. This branch is rebased enough against `master` that you shouldn't have any issues building it from source. In fact, the has the command that can be used to build [the standard binary](https://gist.github.com/anonymous/d608a617a8c984a6ffe273ef2f989a1c) as well as [other builds](https://travis-ci.org/codedellemc/rexray/builds/272558201).

benoitm76:
Is this pull request fixing the wrong Volume Path bug of Cinder ?

Indeed there is a known bug where Cinder Api returning the wrong VolumePath for an attached volume in some cases : https://bugs.launchpad.net/nova/+bug/1217874

For example, Cinder API says that my volume X is attach to `/dev/vdd` but in reality on my instance it is attached on path `/dev/vdc`.

akutz:
Hi @benoitm76,

I'll let @MatMaul answer. I *could* tell you, but I don't want to seem like a know-it-all bigshot, and it's really Mat's baby, so he should explain. I mean, I totally *could* explain it. No lie. Definitely I could do so...but yeah, Mat should 

:)

MatMaul:
@akutz :) @benoitm76 I don't remember dealing explicitly with that so probably not.
A new version has been released so I'll let you check the state and we can have a look if it is still the case.

PR #991, opened 2017-08-28
Simplified Structure

akutz:
This patch accomplishes the following:

* Simplifies the source tree structure
* Simplifies the build tags
* Removes the ability to start service in the background. The service is always started in the foreground and relies on SystemD to start in the background.
* `-l,--logLevel` happens before everything
* There is an `agent` package at the root of RR; this is the service responsible for hosting long running modules
* There are no more `libstorage` file or directory names expected or created at runtime. Everything happens under the brand `rexray`
* The REX-Ray `util` package now defers to libStorage's path config. This means that `rexray start` as a normal user will automatically create files in your `$HOME/.rexray` directory if you don't have root permissions.
* `-c,--config` specifies a custom config and all other config *files* are ignored


libstorage/api/context/context.go
```diff
@@ -1,14 +1,14 @@
 package context
 
 import (
+	"context"
```
codenrhoden: so is this possible because we are only on GO 1.[7-9] now? If so, do we need to update `travis.yml` to reflect that?

akutz: Hi @codenrhoden,

The `.travis.yml` Go version is 1.8.3 as we discussed.

PR #993, opened 2017-08-29
Dep Support

akutz:
This patch introduces support for the official Go dependency management tool, [Dep](https://github.com/golang/dep). Glide support has been removed in favor of the new, official tool.


akutz:
Hi @codenrhoden,

> The pedant in me would want to see this split into two commits. One that replaces Glide with the same dependency versions (or as close to it as we can get) so we can see what changes in vendor/. Ideally switching it out would have minimal changes to vendor/. Then, a second commit that upgrades dependencies where we've chosen to do so.

So I *did* attempt to do this. The PR's commit should not introduce any new or updated versions of dependencies. The changes you see are due to the fact that I did not attempt to recreate a `Gopkg.toml` that is equivalent to the old `glide.yaml` file with explicit, transitive dependencies. You even mention this:

> I think the hardest part in achieving that would be that we are going to back to having `Gopkg.toml` only list immediate dependencies, not transitive ones.

I'm going to spend 30 or so minutes to see if I am able to create a `Gopkg.toml` that is equivalent to the deprecated `glide.yaml`, transitive dependencies and all. At least then we should see an initial `diff` with the only changes being the manifest files and not `vendor/`. Mostly anyway. There is also an issue with the [`go-yaml`](https://twitter.com/ssakutz/status/902638605679681537) dependency.

akutz:
Hi @codenrhoden,

I'm going to add a second commit to this PR that updates the versions of the dependencies by virtue of undoing the overrides in `Gopkg.toml`. The reason I'm adding the commit to this same PR is so we can easily revert the PR in order to back out *all* these changes if Dep + stripping nested vendor breaks anything.

codenrhoden:
Awesome. I really like being able to see what versions changed between the two commits. And having the transitive deps out of the `.toml` file is a big win.

Scrolling through all the deps, I think the only thing I noticed is that we don't have Cobra as a direct dependency, so it gets pointed to `master`, which is a little surprising. Maybe something to think about locking down in the future.

akutz:
Hi @codenrhoden,

I force pushed this PR with the second commit updated to reflect Cobra and PFlag as direct entries in the `Gopkg.toml` file per your suggestion. The reason for their absence in the first place is @spf13 and his project maintainers' decision to forego tagged releases. Without tagged releases the Cobra and PFlag dependencies simply point at `master` unless we lock them to a specific revision.

PR #995, opened 2017-08-31
Sirupsen to sirupsen

akutz:
This patch handles the chaos caused when the Github user account Sirupsen was renamed to sirupsen.


akutz:
Hi @codenrhoden,

I was working for a while trying to undo some of this PR and instead update only the dependencies necessary to solve the logrus issue. However, golang/dep#1104 partially explains why this isn't possible. Due to the way Dep solves constraints, `dep ensure -update github.com/akutz/gofig` fails because the entire resolution must occur before updating only `gofig`. However, the resolution fails because of the original logrus issue.

Issue #904, opened 2017-06-27
Create REX-Ray module for CSI Ident+Node+Controller Services


codenrhoden:
# Summary
For initial CSI support, create a new REX-Ray module (or modules) that implement the CSI Identity, Node, and Controller services.

# New Feature
The module(s) listens on the given `CSI_ENDPOINT` for gRPC traffic, and translates that to our existing libStorage API.

The module (or modules) need to be configurable such that we can create the following "CSI plugins":

* node -- this plugin serves the Node and Identity RPCs
* controller -- this plugin serves the Controller and Identity RPCs
* node+controller -- this plugins serves all Node, Controller, and Identity RPCs

akutz:
Hi @clintkitson / @codenrhoden / et. al,

Some topics I wish to discuss regarding the new CSI module:

## Existing Integration Workflow
What are your thoughts on how the Northbound Adapter (thanks @koensayr) for CSI will function with regards to the existing integration workflow. Should requests handled by the CSI module be handled by the same integration driver used by the Docker module? If so this means:

* Volumes created by the CSI module a root `data` directory.
* The CSI module can be configured with the same [Volume](http://libstorage.readthedocs.io/en/stable/user-guide/config/#volume-configuration) configuration properties that the Docker module can use.

## Default Module Configuration
Currently the CSI module has no analogue to a `default-docker` configuration, ex. `default-csi`. This is because the CSI specification does not define a default location for UNIX socket files. Should REX-Ray define a default location for a CSI socket file and in turn define a `default-csi` module that is configured if the REX-Ray standard or agent binaries run as services?

I personally think that REX-Ray could also drop the `default-docker` module configuration as well and move towards requiring explicit configurations for modules, but that would break backwards compatibility, so that's just an aside.

clintkitson:
The `data` directory is intentionally not declared in the CSI specification. The workflow is somewhat different where the CO tells the plugin where to eventually get something mounted. I believe we would still use the `data` directory and do an additionally bind mount to the CO defined location from this `data` directory.

In terms of the integration driver, I don't think I have enough information one way or another. Obviously today it is an unused abstraction since there is only one driver. I think the priority is to ensure existing Docker/Mesos/Flex functionality works as is alongside the new CSI functionality.

clintkitson:
Doesn't the CSI spec declare that the CO passes a socket location for the plugin where it should then place a socket file?

akutz:
Hi @clintkitson,

[...]

PR #998, opened 2017-09-05
CSI Publish/Unpublish

codenrhoden:
putting this up for early review.

Not even tested yet.

### DIFF
--- a/agent/csi/libstorage/csi_service_libstorage.go
+++ b/agent/csi/libstorage/csi_service_libstorage.go
@@ -2,25 +2,35 @@ package libstorage
 
 import (
 	"context"
+	"fmt"
 	"math"
 	"net"
+	"os"
+	"path"
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"google.golang.org/grpc"
 
-	log "github.com/sirupsen/logrus"
 	xctx "golang.org/x/net/context"
 
 	gofig "github.com/akutz/gofig/types"
 	"github.com/codedellemc/gocsi"
 	"github.com/codedellemc/gocsi/csi"
+	"github.com/codedellemc/gocsi/mount"
 	"github.com/codedellemc/goioc"
 
 	apictx "github.com/codedellemc/rexray/libstorage/api/context"
 	apitypes "github.com/codedellemc/rexray/libstorage/api/types"
 	apiutils "github.com/codedellemc/rexray/libstorage/api/utils"
+)
+
+const (
+	// WFDTimeout is the number of seconds to set the timeout to when
+	// calling WaitForDevice
+	WFDTimeout = 10
 )
 
 func init() {
@@ -30,6 +40,8 @@ func init() {
 // ctxConfigKey is an interface-wrapped key used to access a possible
 // config object in the context given to the provider's Serve function
 var ctxConfigKey = interface{}("csi.config")
+
+var ctxExactMountKey = interface{}("exactmount")
 
 type driver struct {
 	ctx          apitypes.Context
@@ -53,7 +65,7 @@ func (d *driver) Serve(ctx context.Context, lis net.Listener) error {
 
 	// Check for a gofig.Config in the context.
 	if config, ok := d.ctx.Value(ctxConfigKey).(gofig.Config); ok {
-		log.Info("init csi libstorage bridge w ctx.config")
+		d.ctx.Info("init csi libstorage bridge w ctx.config")
 		d.config = config
 	}
 
@@ -204,6 +216,13 @@ func (d *driver) ControllerPublishVolume(
 	req *csi.ControllerPublishVolumeRequest) (
 	*csi.ControllerPublishVolumeResponse, error) {
 
+	volumeID, ok := req.VolumeId.Values["id"]
+	if !ok {
+		return gocsi.ErrControllerPublishVolume(
+			csi.Error_ControllerPublishVolumeError_INVALID_VOLUME_ID,
+			`missing "id" field`), nil
+	}
+
 	// Make sure the requested volume capability is supported for
 	// the cached storage type.
 	if !isVolCapSupported(d.storType, req.VolumeCapability) {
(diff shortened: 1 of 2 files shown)


agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -192,25 +204,72 @@ func (d *driver) ControllerPublishVolume(
...
 			err.Error()), nil
 	}
-	_ = iid
 
-	//d.client.Storage().VolumeAttach(d.ctx, volumeID, opts)
+	opts := &apitypes.VolumeAttachOpts{Opts: apiutils.NewStore()}
 
-	return nil, nil
+	// Get next device via Executor
+	nd, err := d.client.Executor().NextDevice(d.ctx, apiutils.NewStore())
```
akutz: Hi @codenrhoden,

Not all storage drivers require the next device logic, and anyway, it's not needed at all since the libStorage `libstorage` storage driver that is used by the libStorage client handles that for you:

```go
func (d *driver) VolumeAttach(
	ctx types.Context,
	volumeID string,
	opts *types.VolumeAttachOpts) (*types.Volume, string, error) {

	if d.isController() {
		return nil, "", utils.NewUnsupportedForClientTypeError(
			d.clientType, "VolumeAttach")
	}

	ctx = d.requireCtx(ctx)
	serviceName, ok := context.ServiceName(ctx)
	if !ok {
		return nil, "", goof.New("missing service name")
	}

	nextDevice, err := d.NextDevice(ctx, utils.NewStore())
	if err != nil {
		return nil, "", err
	}

	var nextDevicePtr *string
	if nextDevice != "" {
		nextDevicePtr = &nextDevice
	}

	req := &types.VolumeAttachRequest{
		NextDeviceName: nextDevicePtr,
		Force:          opts.Force,
		Opts:           opts.Opts.Map(),
	}

	return d.client.VolumeAttach(ctx, serviceName, volumeID, req)
}
```

All you should need to do is use `Storage().VolumeAttach`.

agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -361,15 +429,141 @@ func (d *driver) NodePublishVolume(
 			"invalid volume capability request"), nil
 	}
 
-	return nil, nil
+	if mv := req.VolumeCapability.GetMount(); mv != nil {
```
akutz: Hi @codenrhoden,

Might I suggest refactoring this and the next, long, indented logic block into two functions? It may help with debugging in the future and certainly with readability.

codenrhoden: Not a problem. That's actually how I have it set up in `csi-blockdevices`, so it's a really easy change.

codenrhoden:
Okay, new version is up.  I think I need to do one more rebase.

This fills out the logic for Controller Publish/Unpublish, and Node
Publish/Unpublish. The controller Publish uses the libStorage
VolumeAttach() logic, which merely attaches the block device to the
given node and then returns the device token in the PublishVolumeInfo.
This is important as that token needs to be provided for Node Publish
operations.

Controller Unpublish does the reverse, and detaches the volume from the
node.

Node publish uses the existing integration driver's Mount logic, which
will mount the device in the standard /var/lib/rexray/volumes/...
location, or wherever it is configured to do so. It also adds the /data
directory into the volume when it is mounted. The /data dir is
bind-mounted to the taret.  Node publish also will bind-mount the
underlying device, specificied by the PublishVolumeInfo token, directly
to a file specified by target if BlockVolume is used instead.

Node Unpublish removes the bind mount, and the private mount in
/var/lib/rexray if the volume is not mounted anywhere else.

agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -381,15 +445,94 @@ func (d *driver) NodePublishVolume(
...
+
+	target := req.GetTargetPath()
+
+	opts := &apitypes.VolumeInspectOpts{Opts: apiutils.NewStore()}
+	opts.Attachments = apitypes.VolAttReqWithDevMapOnlyVolsAttachedToInstance
+	vol, err := d.client.Storage().VolumeInspect(d.ctx, volumeID, opts)
+	if err != nil {
+		return nil, err
+	}
+	if vol == nil {
```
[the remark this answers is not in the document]
codenrhoden: Although this is something that needs to be fixed, as I just tried it out (hadn't done that) where I call a `NodeUnpublish` *after* a succesful `ControllerUnpublish`, which is something that shouldn't happen per the spec. But if you do, this logic fails because you get an HTTP 404 not found from libStorage, and we need to check for that -- it is an expected error.

agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -453,3 +596,144 @@ func (d *driver) NodeGetCapabilities(
...
+
+	d.ctx.WithField("path", target).Debug(
+		"checking if private mount can be unmounted")
+
+	mnts, err := mount.GetDevMounts(device)
+	if err != nil {
+		return err
+	}
+
+	d.ctx.Debugf("%+v", mnts)
```
codenrhoden: ack, I need to remove this.

libstorage/drivers/integration/linux/linux.go
```diff
@@ -273,13 +275,28 @@ func (d *driver) Mount(
...
+		if _, ok := ctx.Value(ctxExactMountKey).(interface{}); ok {
+			for _, m := range mounts {
+				if m.MountPoint == mountPath {
+					ctx.Debug("returning existing mount")
+					return d.volumeMountPath(
+						m.MountPoint), vol, nil
+				}
+			}
+			return "", nil, goof.New("device is already mounted")
+		}
```
codenrhoden: This code previously just returned the first mount it saw with the matching device name. That's not good enough when we are using bind mounts, as there will be multiple entries. That's why the "exactmatch" context was added -- we only return the mountpath for a mount matching both the device and the expected mountpoint.

For example, if the mount table had this:
```
/dev/xvdf /mnt/test xfs rw,seclabel,relatime,nouuid,attr2,inode64,noquota 0 0
/dev/xvdf /var/lib/rexray/volumes/test/data xfs rw,seclabel,relatime,nouuid,attr2,inode64,noquota 0 0
```
We would return `/mnt/test/data` for the mount path, when we really wanted `/var/lib/rexray/volumes/test/data`

codenrhoden:
Okay, I added the WaitForDevices logic in `NodePublishVolume`.

Closes #904

PR #1003, opened 2017-09-06
Add csi-blockdevices driver

### LINKED ISSUE #983: Integrate csi-blockdevices driver with REX-Ray

codenrhoden:
Closes #983


codenrhoden:
One odd thing with the `dep ensure` here to me was that there were changes in `vendor/` that did not seem accounted for by the changes in `Gopkg.lock`. I'm not sure why this is, unless new things had been added to `vendor/` before by `dep ensure`, and those new things were added to the commit.

Otherwise, if a commit hash for a dependency hasn't changed, I would not expect to see changes related to those projects in `/vendor`.

akutz:
Replaced by #1005.

PR #1008, opened 2017-09-07
Idempotent interceptor fixes

codenrhoden:
This PR has two commits.

The first fixes an issue of the interceptor returning an error when the device is not found in the mount table -- this is not an error, it just means it hasn't been mounted (and the interceptor should allow the logic to continue to the CSI bridge.

The second commit adresses a couple of things. First is to account for the `/data` folder that occurs in libStorage mounted filesystems.

The second thing is to use the device token from the PublishVolumeInfo to use `LocalDevices` to get the underlying device when it is present. During an unpublish, this pointer is `nil`, and we have to fall back to a `VolumeInspect`.

As I am writing this, it dawned on me that the logic may still have issues with `BlockVolume` publishes, as right now it is blindly assuming it's a file system and appending `/data`. That's not going to work for `BlockVolume`... I see another change coming...

I tested in this on EBS, with doing attach, mount, unmount, detach repeatedly. Secondary calls were issued for all, and the idempotent interceptor picked them up correctly.

An area that needs to be tested is covering the bases for all types of storage/local-devices. Namely, EBS that returns a token of an actual device, ScaleIO that returns a token that get's transformed into a device, and EFS/S3FS that return no token at all.

### DIFF
--- a/agent/csi/libstorage/csi_service_libstorage_idemp.go
+++ b/agent/csi/libstorage/csi_service_libstorage_idemp.go
@@ -3,6 +3,7 @@ package libstorage
 import (
 	"errors"
 	"fmt"
+	"path"
 
 	"github.com/codedellemc/gocsi/csi"
 	"github.com/codedellemc/gocsi/mount"
@@ -12,6 +13,8 @@ import (
 )
 
 var errMissingIDKeyPath = errors.New("missing id key path")
+var errMissingTokenKey = errors.New("missing token key")
+var errUnableToGetLocDevs = errors.New("unable to get local devices")
 
 const resNotFound = "resource not found"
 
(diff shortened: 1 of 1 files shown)


agent/csi/libstorage/csi_service_libstorage_idemp.go
```diff
@@ -122,33 +124,69 @@ func (d *driver) IsNodePublished(
...
-		return false, err
-	}
+		// Get device from local devices
+		opts := &apitypes.LocalDevicesOpts{
+			Opts:     apiutils.NewStore(),
+			ScanType: apitypes.DeviceScanQuick,
+		}
+		devs, err := d.client.Executor().LocalDevices(d.ctx, opts)
+		if err != nil {
+			return false, fmt.Errorf("unable to get local devices")
```
akutz: Hi @codenrhoden,

Please consider creating a package field like so:

```
var errUnableToGetLocDevs = errors.New("unable to get local devices")
```

This way a new error isn't allocated each time there is a failure.

PR #1009, opened 2017-09-08
Idempotent Errors

akutz:
This patch updates the idempotent interceptor to allow for the cases where the guarded operation returns an error. If it does, that call for that volume will be marked as in error and a subsequent, same call for that volume will bypass the idempotent guard. Not until the handler returns a successful operation will the idempotent guard take effect once again for that volume for that particular operation.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -92,16 +92,16 @@
   revision = "32ae09ae44241977ba9f30bb2ae81bf8d8b2aa1c"
 
 [[projects]]
+  branch = "master"
   name = "github.com/codedellemc/csi-vfs"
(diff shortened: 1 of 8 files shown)


akutz:
Hi @codenrhoden,

Nope. Just a made up property. However, `time.ParseDuration`, when errors, returns a duration of `0s` because Duration is a primitive and cannot have a nil zero value. Thus its zero value, is, well, zero. And that was the previous timeout used by the `TryLock` function in the interceptor. I just made it configurable outside the interceptor in case we ever want to determine how much time to wait to obtain a lock.

PR #1011, opened 2017-09-08
Update idemp provider to handle block and mnt vols

codenrhoden:
This patch updates the idempotency provider to handle both block and
mount volume types, as they need to be handled a bit differently. This
information is not always given directly in the CSI request, so right
now it is inferred by the target path. If the path is a dir, we assume
mount, if the path is a file, we assume block.

We also simplify the logic of looking for mounts (to see if published or
not) by only matching the source device and the target, rather than also
caring about the intermediate "private" mount in between. We can do this
because the interceptor has been been updated previously to catch any
errors returned from the CSI bridge, and if there are any "dangling"
mounts, we trust our bridge to return an error, which should cause a
retry.

We also updated the CSI bridge to behave directly when the idempotency
interceptor and provider are not in play -- the logic to look for
existing mounts was not in place, but is now.

### DIFF
--- a/agent/csi/libstorage/csi_service_libstorage.go
+++ b/agent/csi/libstorage/csi_service_libstorage.go
@@ -2,6 +2,7 @@ package libstorage
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"net"
@@ -38,28 +39,34 @@ var (
 	// config object in the context given to the provider's Serve function
 	ctxConfigKey     = interface{}("csi.config")
 	ctxExactMountKey = interface{}("exactmount")
+	errDirNeeded     = errors.New("target path needs to be a directory")
+	errFileNeeded    = errors.New("target path needs to be a file")
 )
+
+type pubInfoCache struct {
+	attToken string
+}
 
 func init() {
 	goioc.Register("libstorage", func() interface{} { return &driver{} })
 }
 
 type driver struct {
-	ctx          apitypes.Context
-	client       apitypes.Client
-	config       gofig.Config
-	server       *grpc.Server
-	svcName      string
-	storType     apitypes.StorageType
-	iid          *apitypes.InstanceID
-	nodeID       *csi.NodeID
-	attTokens    map[string]string
-	attTokensRWL sync.RWMutex
+	ctx        apitypes.Context
+	client     apitypes.Client
+	config     gofig.Config
+	server     *grpc.Server
+	svcName    string
+	storType   apitypes.StorageType
+	iid        *apitypes.InstanceID
+	nodeID     *csi.NodeID
+	pubInfo    map[string]*pubInfoCache
+	pubInfoRWL sync.RWMutex
 }
 
 func (d *driver) Serve(ctx context.Context, lis net.Listener) error {
 
-	d.attTokens = map[string]string{}
+	d.pubInfo = map[string]*pubInfoCache{}
 
 	d.ctx = apictx.New(ctx)
 	d.client = apictx.MustClient(d.ctx)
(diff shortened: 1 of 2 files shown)


codenrhoden:
@akutz this patch also kept the changing of `attTokens` over to a struct, even though it is not necessary. Let me know if you are okay with that.

agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -292,6 +299,10 @@ func (d *driver) ControllerUnpublishVolume(
 		return nil, err
 	}
 
+	d.pubInfoRWL.Lock()
+	defer d.pubInfoRWL.Unlock()
+	delete(map[string]*pubInfoCache{}, volumeID)
```
akutz: Hi @codenrhoden,

Shouldn't this line be:

```go
delete(d.pubInfo, volumeID)
```

agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -448,6 +459,11 @@ func (d *driver) NodePublishVolume(
 	target := req.GetTargetPath()
 	vi := req.GetPublishVolumeInfo()
 
+	st, err := os.Stat(target)
```
akutz: Hi @codenrhoden,

We already import gotil, so why not use it? Just curious.

agent/csi/libstorage/csi_service_libstorage.go
```diff
@@ -486,10 +502,16 @@ func (d *driver) NodePublishVolume(
 	}
 
 	if mv := req.VolumeCapability.GetMount(); mv != nil {
+		if !st.IsDir() {
```
akutz: Hi @codenrhoden,

Nevermind, I see why you didn't use gotil. You're using the result of the stat call here. Thanks.

akutz:
Hi @codenrhoden,

Because of the required change to the `delete` call, if you want to go ahead and change back the attach token cache from structs to bools, that's fine. Otherwise I wouldn't bother. But since a change is required anyway...

agent/csi/libstorage/csi_service_libstorage.go
codenrhoden: probably. I think I got that from Atom's autofill.  Sad, but I've never had to delete an item from a map before. :)  I can make that change.

agent/csi/libstorage/csi_service_libstorage.go
codenrhoden: yep, exactly. I started with `gotil` as requested, but then noticed I was going to use `os.Stat` to find out if it is a dir anyways...

codenrhoden:
@akutz, yeah, I could go either way on the token struct. I'm inclined to leave it, as it allows for other things to be added if needed. But it is also unlikely to be needed, as that would imply some addition to libStorage across all drivers...

I'll leave it unless you say no. :)

agent/csi/libstorage/csi_service_libstorage.go
akutz: Let me restate it -- My suggestion wasn't really a suggestion. I was being polite :)

akutz:
Hi @codenrhoden,

I'm with you -- let's leave the cache in case we need it for something else in the future.

agent/csi/libstorage/csi_service_libstorage.go
codenrhoden: thanks for catching this. silly mistake

akutz: No worries. Good catch on the Atom autofill doing the wrong thing. That's weird. And I guess I see why no linter or compiler caught it. You're technically deleting a key/value pair from a map. It's just a brand new map :)

PR #1031, opened 2017-09-14
Add GCE Convert Underscores Configuration.

jamespedwards42:
Fixes https://github.com/codedellemc/rexray/issues/1020

@codenrhoden I just noticed https://github.com/codedellemc/rexray/pull/1030 as I created this PR, went ahead and added this one as well for the documentation updates as well.

Has the default value of `false` for converting underscores been discussed before for Digital Ocean and GCE?  It feels more reasonable to default to `true` given that it won't work otherwise... Or is the idea that we want it to break so that the user notices what is happening?

### DIFF
--- a/.docker/plugins/gcepd/config.json
+++ b/.docker/plugins/gcepd/config.json
@@ -37,6 +37,14 @@
         },
         {
           "Description": "",
+          "Name": "GCEPD_CONVERTUNDERSCORES",
+          "Settable": [
+            "value"
+          ],
+          "Value": "false"
+        },
+        {
+          "Description": "",
           "Name": "GCEPD_DEFAULTDISKTYPE",
           "Settable": [
             "value"
--- a/.docs/user-guide/docker-plugins.md
+++ b/.docs/user-guide/docker-plugins.md
@@ -348,6 +348,11 @@ $ docker plugin install rexray/gcepd \
 The GCEPD plug-in requires that GCE compute instance has Read/Write Cloud API
 access to the Compute Engine and Storage services.
 
+**NOTE:** GCE persistent disks cannot be created if their name contains an underscore.
+Docker will automatically append prefixes with underscores to your volume names when
+they are created as part of a compose file, so if you're creating volumes with this plugin
+using compose (or stack deploy), be sure to set `GCEPD_CONVERTUNDERSCORES` to `true`.
+
 #### Privileges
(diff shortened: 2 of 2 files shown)


akutz:
Hi @jamespedwards42,

I think you raise a terrific question regarding the default value. However, I don't see any reason to delay merging this PR until such a discussion -- the default value and associated documentation can always be updated. At least this way there *is* documentation -- thank you for providing it!

codenrhoden:
Hi @jamespedwards42 

> Has the default value of false for converting underscores been discussed before for Digital Ocean and GCE? It feels more reasonable to default to true given that it won't work otherwise... Or is the idea that we want it to break so that the user notices what is happening?

It hadn't really been discussed, it was just what I chose to do at the time. My reasoning there was that this was a bit of magic, and it has some potential for side effects, so right now it is a choice to turn it on.

With the option enabled, we are doing something slightly different than what is asked. While necessary for stack/swarm/compose, it bothers me all the same. It also relies on behavior that could very well change in Docker. Docker does not verify that the volume it asked for is what it gets back -- for example, `docker volume inspect project_volume` will happily report and use a volume named `project-volume`, which is the intent, but if it ever bothered to look at the response it gets back, it would see that the volume name returned does not match volume name requested. I don't think this behavior is going to change, but again, it bothers me.

But, I am happy that someone is actually looking at it and finding the option useful. A lot of internal work went into making this seemingly simple feature possible. If enabling the option by default in the Docker managed plugin is something people want, we are happy to consider it. It makes sense from a usability perspective - I'm probably just paranoid. Plus, we have the option of making it default to true for the plugin (when we know it is being used by Docker) and false outside the plugin (when, from a driver perspective, we don't particularly know/care if being used by Docker).

jamespedwards42:
> A lot of internal work went into making this seemingly simple feature possible.

I can imagine, I spent an hours trying to figure how the environment variable was magically linked to some actual logic :) ... I hope users appreciate how great Rexray is and how incredibly valuable it is, I certainly do.

I went digging to see what the current view of underscore name spacing is for compose.  It looks like it is causing headaches for other users as well, particularly for hostnames (https://github.com/docker/compose/issues/4475). 

I agree with your view that it is probably not worth the risk of breaking future docker volume name checks.  Maybe this issue with DO and GCE will help put pressure on the compose project to fix it properly by allowing the delimiter to be configured.

Issue #957, opened 2017-08-07
config option for RBD object size


codenrhoden:
@codenrhoden commented on [Tue Jan 03 2017](https://github.com/codedellemc/libstorage/issues/381)

The RBD driver currently uses the default 4MB object size when creating new RBDs. Add a config option to specify difference object sizes.

---

@richard-julien commented on [Thu Mar 02 2017](https://github.com/codedellemc/libstorage/issues/381#issuecomment-283751071)

Hi, I just started to use libStorage with Ceph and for me some others options will be very intesting. (like --cluster, -m, --keyring, ...)
Do you think we could transform this tickets to support various options, and maybe the simplest way is to support a extra_options where the user can setup everything he wants instead of mapping the rdb command field by field?
I'm new to go and libStorage but I can try to help on this ;)

---

@codenrhoden commented on [Wed Mar 22 2017](https://github.com/codedellemc/libstorage/issues/381#issuecomment-288557935)

@richard-julien That's a really interesting idea, the `extra_options` proposal. I'll think about it. The first thing that came to mind was wondering if you could use `CEPH_ARGS` today and accomplish some of what you need. That's an env var that the Ceph CLI tools look for already, but it's been a while since I've used it.

Second thought was wondering how to de-conflict if we do end up adding first-class support for any of these options (say `rbd.cluster`) and you also specify `--cluster` in the `extra_options` field.

Anyways, it's good to know someone is looking for this functionality. That helps us prioritize what to work on!

---

@richard-julien commented on [Wed Mar 29 2017](https://github.com/codedellemc/libstorage/issues/381#issuecomment-290032384)

Hi @codenrhoden thanks for your answer.
My idea is to be able to configure a ceph mount without any specific configuration file on the host. (everything through the volume-opt). So the CEPH_ARGS for me is not really an option ;)

For the extra_options and the de-conflict maybe the simplest solution for the ceph driver is to only provide a free text zone. I dont think libStorage need to ensure the consistency of the command, and helping the user with 2 or 3 options is not really necessary. I will be happy to provide every options i need in the command line and have directly the ceph error message if i forget to provide the basic options.

PR #1035, opened 2017-09-18
rbd: Add cephArgs config option

### LINKED ISSUE #924: rbd: Configurable cephx keyring
# Summary
rbd tries to use client.admin by default, and the plugin does not appear to accept any of the parameters that could alter this (for instance, CEPH_ARGS)

# New Feature
It would be helpful if one could configure the cephx user, for instance by simply allowing passing CEPH_ARGS.

## Version
rexray/rbd:9.2

### LINKED ISSUE #955: rbd: Allow non-admin user
@codenrhoden commented on [Tue Jan 03 2017](https://github.com/codedellemc/libstorage/issues/379)

Right now the driver only uses the default `client.admin` user. Allow a config option to specify a different user.

An open question is whether we should all the user's cephx key to be stored in our config file, or whether we require it to be configured in `ceph.conf`. I'm leaning toward the later. As long as we require a `ceph.conf` to be present, we should push as much configuration into it as we can, rather than proliferate sensitive credentials into new/additional locations.

codenrhoden:
This patch adds an rbd.cephArgs config option. It takes free form text,
and is used to set the CEPH_ARGS env var. This env var is mostly used to
set "--id" and "--cluster" CLI args, the former being the name client
user to use (instead of admin) and the latter for using a non-default
cluster name (default is Ceph).

Using this field, a non-admin client and keyring can be used. Using that
client successfully means that the keyring has to be present in
/etc/ceph, and that the client's auth caps are set correctly. It is easy
to get in a situation where those are wrong, and calls to "rbd" or
"rados" hang instead of issuing "operation not permitted". To help
combat this, the driver no longer supports multiple pools when cephArgs
is set. Only the single pool defined as the defaultPool is supported.

I would have removed multi-pool support entirely, but that would break
backwards compatibility, and I've seen users use this functionality. But
setting CephArgs limits you to one pool only.

Replaces #934 
Fixes #924 
Fixes #955
xref #957 (addresses parts of it)

This patch also has an unrelated, 2nd commit with a fix I found necessary while working on the new feature.

Calling cmd.Help() prints the help output, but it does not exit the
program. There were places in the code were would detect a critical
error, emit it stderr, then emit the help output, but *not* exit the
program. This caused the program to continue execution leading to
potential faults from nil dereferences depending on what went wrong. We
really just want to emit the help then shutdown.

### DIFF
--- a/.docker/plugins/rbd/config.json
+++ b/.docker/plugins/rbd/config.json
@@ -53,6 +53,14 @@
         },
         {
           "Description": "",
+          "Name": "RBD_CEPHARGS",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "",
           "Name": "RBD_DEFAULTPOOL",
           "Settable": [
             "value"
--- a/.docs/user-guide/docker-plugins.md
+++ b/.docs/user-guide/docker-plugins.md
@@ -205,6 +205,7 @@ plug-in:
 
 Environment Variable | Description | Default | Required
 ---------------------|-------------|---------|---------
+`RBD_CEPHARGS` | Text to set in the `CEPH_ARGS` environment variable | ""
 `RBD_DEFAULTPOOL` | Default Ceph pool for volumes | `rbd`
 
 ## Dell EMC
--- a/.docs/user-guide/storage-providers.md
+++ b/.docs/user-guide/storage-providers.md
@@ -407,9 +407,10 @@ cluster.
 * The `ceph` and `rbd` binary executables must be installed on the host
 * The minimum required version of Ceph *clients* is Infernalis, `v9.2.1`
 * The `rbd` kernel module must be installed
-* A `ceph.conf` file must be present in its default location
-  (`/etc/ceph/ceph.conf`)
-* The ceph `admin` key must be present in `/etc/ceph/`
+* A `<clustername>.conf` file must be present in its default location
+  (`/etc/ceph/`). By default, this is `/etc/ceph/ceph.conf`.
+* The specified user (`admin` by default) must have a keyring present in
+  `/etc/ceph/`.
 
 #### Configuration
 The following is an example with all possible fields configured. For a running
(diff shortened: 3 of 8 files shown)


clintkitson:
@codenrhoden Are we exposing this to the docker managed plugin as well?

codenrhoden:
@clintkitson It's not there yet, but it absolutely should be.

akutz:
Hi @codenrhoden,

Should we merge this PR or let you first expose the above env vars to the Docker plug-in?

codenrhoden:
@akutz no merge yet. When I asked @mvollman to review this, the comments ended up on his PR, #934.  I have some feedback in there to address still.

codenrhoden:
@mvollman okay, I've relaxed the previous restriction a bit, such that it still allows multi pool if you are only setting something like `--cluster` in CEPH_ARGS. Basically, since we default to the admin user, if you don't do something to change it to something else, we know it's still using the admin user. But if you use any of the flags to set the user, we don't know what is going to happen, so we take the safest route of only allowing you the pool you set.

A future enhancement could indeed be to allow a list of pools. I'd be happy if you could take a look.

akutz:
HI @codenrhoden,

Please rebase this since I merged the doc change.

mvollman:
@codenrhoden I was waiting to review this once it was finished and then I just saw that it was merged.  I see that you have added cephArgs, but I do not see that you are using them.  I built and tested the code and I am not able to use a non-admin user.  Your rbd, rados and ceph commands need to be passed the cephArgs.

codenrhoden:
@mvollman The `CEPH_ARGS` env var gets set here: https://github.com/codedellemc/rexray/blob/a35e8bba97a94f427dd9da965332e220aee22d4c/libstorage/drivers/storage/rbd/utils/utils.go#L378-L380

in my tests, I was able to use both a non-admin user, and use a custom cluster name.

codenrhoden:
@mvollman FWIW, there is an RC being created right now, and once that is up I'd be thrilled if you can install it and give it a try. That way we are sure of the version.

PR #1039, opened 2017-09-22
cinder: Detach volume from all attached nodes

### LINKED ISSUE #913: cinder:  Volume detection & detach fails w Docker Swarm
# Summary
When I create stateful services in swarm mode and they fail, or a node goes down and the service gets rescheduled, the stateful services keep on having cinder errors. This makes it a manual process to remove each of these services manually unmount the volumes with cinder or openstack panel, and redeploy the stacks. If there is a sudden failure, this would effectively defeat the purpose of swarm orchestration...

# Bug Reports
Cinder driver fails to detach or detect volume in docker swarm mode. I currently have settings like recommended [here](https://github.com/codedellemc/rexray/issues/861#issuecomment-309980869).

## Version

```shell
$ rexray version
REX-Ray
-------  
Binary: /usr/bin/rexray
SemVer: 0.9.1  
OsArch: Linux-x86_64
Branch: v0.9.1
Commit: 2373541479478b817a8b143629e552f404f75226  
Formed: Fri, 09 Jun 2017 19:23:38 UTC
  
libStorage
----------  
SemVer: 0.6.1
OsArch: Linux-x86_64  
Branch: v0.9.1
Commit: fd26f0ec72b077ffa7c82160fbd12a276e12c2ad
Formed: Fri, 09 Jun 2017 19:23:05 UTC
```

## Expected Behavior
The failed services should restart with the appropriate volumes mounted without any errors.

## Actual Behavior
Getting either:
VolumeDriver.Mount: {"Error":"error detaching volume"}
or
VolumeDriver.Mount: {"Error":"open /dev/vdc: no such file or directory"}

codenrhoden:
This patch refactors the Cinder attach/detach logic to properly account
for volume preemption and to make sure that volume attach always
detaches the volume from all nodes.
Fixes #913

@MatMaul I went ahead and took a stab at this. I definitely want you to take a look, and right now it is untested. I will spin up my devstack environment and test pre-emption sometime soon. This should have the behavior I described in my comment in #913. However, it does make one change that is important that I get your input on, which is during the detach it always tries the v2 block storage detach (if it exists) if the previous attempt failed. Previously, it would only try that if `force` was true, but I'm not sure that was the right logic.

### DIFF
--- a/libstorage/drivers/storage/cinder/storage/cinder_storage.go
+++ b/libstorage/drivers/storage/cinder/storage/cinder_storage.go
@@ -23,6 +23,12 @@ import (
 
 const (
 	minSizeGiB = 1
+	hiddenText = "******"
+)
+
+var (
+	errVolAlreadyAttached = goof.New("volume already attached to a host")
+	errVolAlreadyDetached = goof.New("volume already detached")
 )
 
 type driver struct {
@@ -89,12 +95,12 @@ func (d *driver) Init(context types.Context, config gofig.Config) error {
 	if d.password() == "" {
 		fields["password"] = ""
 	} else {
-		fields["password"] = "******"
+		fields["password"] = hiddenText
 	}
 	if d.tokenID() == "" {
 		fields["tokenId"] = ""
 	} else {
-		fields["tokenId"] = "******"
+		fields["tokenId"] = hiddenText
 	}
 	fields["tenantId"] = d.tenantID()
 	fields["tenantName"] = d.tenantName()
@@ -105,7 +111,7 @@ func (d *driver) Init(context types.Context, config gofig.Config) error {
 	if trustID == "" {
 		fields["trustId"] = ""
 	} else {
-		fields["trustId"] = "******"
+		fields["trustId"] = hiddenText
 	}
 
 	d.provider, err = openstack.NewClient(authOpts.IdentityEndpoint)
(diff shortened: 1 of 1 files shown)


MatMaul:
@codenrhoden thanks for that, I'll have a look and test next week.

n1b0r:
Could we help testing ? How could we test ? 

I'm not familiar with go. How could I build this branch to test it out ?

akutz:
Hi @MatMaul,

Have you had an opportunity to validate this fix?

Hi @n1b0r,

You can build this PR locally using Docker with this command:

```bash
$ docker run -it \
  -v "$(pwd)":"/tmp/rexray" \
  golang:1.9 \
  bash -c "git clone https://github.com/codedellemc/rexray \
      \"/go/src/github.com/codedellemc/rexray\" &&
    cd \"/go/src/github.com/codedellemc/rexray\" && 
    git fetch origin +refs/pull/1039/merge: &&
    git fetch --tags origin &&
    git checkout -b a0f0b660 a0f0b660ba1e385dfab4f926c014e3d947632be4 &&
    XGOOS=linux XGOARCH=amd64 GOOS= GOARCH= go generate &&
    GOOS=linux GOARCH=amd64 go build -o \"rexray\" &&
    cp -f \"rexray\" \"/tmp/rexray\"" && \
  md5sum "rexray" && \
  ls -al "rexray"
```

This command comes from this [gist](https://gist.github.com/anonymous/437bc7e42450df048bafceb47aa3f1ce). Read more on how to reproduce REX-Ray PR binaries locally in this [blog post](https://blog.thecodeteam.com/2017/09/13/reproducing-rex-ray/) I wrote :)

PR #1042, opened 2017-09-25
Enable S3FS_ENDPOINT for rexray/s3fs plugin

harshavardhana:
Additionally also enable envs to avoid using "/data"
as ROOTPATH, s3fs plugin doesn't have the same
restrictions as other storage drivers. It is necessary
to make top level prefixes and files to be visible.

Make this a default change such that the container
mounts see the entire bucket contents not just inside
"/data".

### DIFF
--- a/.docker/plugins/s3fs/config.json
+++ b/.docker/plugins/s3fs/config.json
@@ -21,6 +21,30 @@
         },
         {
           "Description": "",
+          "Name": "LIBSTORAGE_INTEGRATION_VOLUME_OPERATIONS_MOUNT_ROOTPATH",
+          "Settable": [
+            "value"
+          ],
+          "Value": "/data"
+        },
+        {
+          "Description": "",
+          "Name": "LINUX_VOLUME_ROOTPATH",
+          "Settable": [
+            "value"
+          ],
+          "Value": "/data"
+        },
+        {
+          "Description": "",
+          "Name": "S3FS_ENDPOINT",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "",
           "Name": "S3FS_ACCESSKEY",
           "Settable": [
             "value"


harshavardhana:
> Thanks for the patch. You've modified the whitespace of the entire file, which makes it hard to see what you have added. I can tell what you've added, but I do ask that the patch not contain whitespace changes. Such changes would also make this config.json file inconsistent with every other plugin.

Sure thing, it is perhaps my editor which auto formatted it @codenrhoden - let me fix it.

akutz:
@codenrhoden,

If you're happy with this please merge it.

codenrhoden:
@akutz.  will do - still waiting on an updated commit that addresses whitespace. Also, needs to be rebased anyways.

codenrhoden:
@akutz @harshavardhana 

I do have one concern about changing the default root path from `/data` to `/`, and that is backwards compatibility. If someone upgrades their plugin to this version, the behavior is going to change, and whatever consumer of this storage volume is going to see completely different data. Previously the would have been inside a `/data` directory, which no ability to get above it in the directory tree, but after this change they would be a directory higher and see a `/data` directory.

I think it may be safer to have the default still be `/data`, and let users change it if they need/want to.  Thoughts?

harshavardhana:
> I think it may be safer to have the default still be /data, and let users change it if they need/want to. Thoughts?

Agreed @codenrhoden

harshavardhana:
PR is updated, sorry for the delay.

codenrhoden:
@harshavardhana Thanks for the update!  This is getting close.  Two things:

- The PR needs to be rebased
- The CI build failed. It looks like it's because of the extra commas.

.docker/plugins/s3fs/config.json
```diff
@@ -21,6 +21,30 @@
         },
         {
           "Description": "",
+          "Name": "LIBSTORAGE_INTEGRATION_VOLUME_OPERATIONS_MOUNT_ROOTPATH",
+          "Settable": [
+            "value"
+          ],
+          "Value": "/data",
```
codenrhoden: The comma here is not necessary, as it's the last item in the map.  Same with the other two.

codenrhoden:
Thanks @harshavardhana !

barrachri:
Hello, does this PR allow to connect to more than one s3_endopint?

harshavardhana:
> Hello, does this PR allow to connect to more than one s3_endpoint?

No it does not @barrachri

clintkitson:
You would do this by creating another instance of the s3fs managed docker
plugin or by using the modules capability of rexray and running in
standalone mode.

clintkitson:
To clarify, are you referring to multiple buckets or endpoints? If
endpoints, what's the use case for that?

barrachri:
This is my use case:

```bash
docker service create \
--name service-with-shared-storage \
--mount type=volume,destination=/path/minio,\
volume-driver=rexray,\
volume-opt=endpoint=http://0.0.0.0:9000,\
volume-opt=access-key=PUTYOURKEY,\
volume-opt=secret-key=PUTYOURSECRETKEY,\
volume-opt=bucket=test-1,\
volume-opt=uid=100 \
ubuntu sleep 60
```
So many services but with 1 endpoint for each (generally).

barrachri:
I would like to help adding this feature.

Basically checking s3-fuse documentation you need to save each s3 credential in a file:

```
echo MYIDENTITY:MYCREDENTIAL > /path/to/passwd
chmod 600 /path/to/passwd
Run s3fs with an existing bucket mybucket and directory /path/to/mountpoint:

s3fs mybucket /path/to/mountpoint -o passwd_file=/path/to/passwd
If you use s3fs with a non-Amazon S3 implementation, specify the URL and path-style requests:

s3fs mybucket /path/to/mountpoint -o passwd_file=/path/to/passwd -o url=http://url.to.s3/ -o use_path_request_style
```
Under the hood rexray is calling s3fs, so an option would be to create a temporary directory inside the plugin container where you create a file named like the volume and put there the credential.

My level of go is the second chapter of flow control (😅) but I would be more than happy to help!

jaredvacanti:
Is it now possible to install the rexray/s3fs plugin on the host machine, and have multiple volumes stored in the same bucket at different mountpoints? Now I am using a new bucket for each docker volume (so a 1-to-1 relationship between replicated services and volumes/buckets). Can I specify a volume to use the same bucket at a different mountpoint?

PR #1055, opened 2017-10-08
docker: Docker-CSI Southbound Bridge

### LINKED ISSUE #1037: csi: Create CSI->Docker southbound bridge
This Epic tracks issues related to creating a new southbound adapter that would translate incoming requests to the docker volume module to a CSI plugin (bypassing libStorage). This means that any CSI plugin could be accessed via Docker, given that CO a way to consume CSI plugins before CSI is supported natively.

akutz:
This patch introduces a Docker-CSI compatibility layer called the Docker-CSI Southbound Bridge. The Docker module has been removed from REX-Ray, and Docker support is now enabled via the CSI module. All CSI module instances will, unless explicitly disabled, multiplex incoming requests to either a CSI service or the Docker Volume API. The Docker Volume API receiver satisfies requests by using a CSI client.

This is a work-in-progress and is not-yet-complete:

- [x] Cursory testing looks good, but still need to complete the detailed testing for each Docker API operation
- [x] Update CSI-VFS dependency to revision [`b386662`](https://github.com/TheCodeTeam/csi-vfs/tree/b386662d75814a056d063082fcb2c596ee7cc9f7)
- [x] Before merging this PR, CSI-VFS should be tagged instead of depending on its revision directly (**tagged as `v0.1.7`**)
- [ ] Tag CSI-NFS and update RR (assigned to @codenrhoden)
- [ ] Tag CSI-BlockDevices and update RR (assigned to @codenrhoden)
- [x] Update the CSI-libStorage bridge
  - [x] Add `name` to `VolumeInfo.Id.Values` returned from the CSI-libStorage bridge
  - [x] Add `targetpaths` to `VolumeInfo.Metadata.Values` returned from the CSI-libStorage bridge. Hopefully @codenrhoden can assist with this
    - [x] Docker bridge sets value in outgoing context that is received by libStorage bridge
    - [x] libStorage bridge looks at above value and returns mount info if value is true
- [x] Perform an initial `List` call to cache the volume name-to-ID mappings
- [x] `Create`
  - [x] Configure sensible defaults for a CSI create call **OR**
  - [x] Parse the Docker options into the CSI `CreateVolume` options
  - [x] `CreateVolume`
  - [x] Add volume to the cache
  - [ ] Better logging
  - [ ] Testing
- [ ] `List`
  - [x] `ListVolumes`
  - [x] Cache `ListVolumes` result for future name-to-ID mappings
  - [ ] Better logging
  - [ ] Testing
- [ ] `Get`
  - [x] Implement
  - [ ] Better logging
  - [ ] Testing
- [ ] `Remove`
  - [x] `NodeUnpublishVolume`
  - [x] `ControllerUnpublishVolume`
  - [x] `DeleteVolume`
  - [x] Remove volume from the cache
  - [ ] Better logging
  - [ ] Testing
- [ ] `Path`
  - [x] Implement
  - [ ] Better logging
  - [ ] Testing
- [ ] `Mount`
  - [x] `GetNodeID`
  - [x] `ControllerPublishVolume`
  - [x] `NodePublishVolume`
  - [x] Update volume's state in the cache
  - [ ] Better logging
  - [ ] Testing
- [ ] `Unmount`
  - [x] `NodeUnpublishVolume`
  - [x] `ControllerUnpublishVolume`
  - [x] Update volume's state in the cache
  - [ ] Better logging
  - [ ] Testing
- [x] `Capabilities`

Fixes #1037


akutz:
cc @clintkitson @codenrhoden @cduchesne 

## The Problem
**NOTE**: This problem does not apply to native CSI plug-ins.

In 0.10.0 we introduced the CSI-libStorage, northbound bridge. This enabled  CSI clients to use all of the storage platforms supported by libStorage. At this  point in time the Docker module in REX-Ray was still accessing libStorage  drivers directly: ex. integration, storage, etc. In other words, this was the model:

```
CSI Client    -> RR CSI Endpoint    -> libStorage bridge -> libStorage drivers

Docker Client -> RR Docker Endpoint -> libStorage drivers
```

### 0.10.x Docker Mount Operation
A Docker `Mount` operation looked like this:

```
Docker Client -> Mount -> RR Docker Endpoint -> libStorage Integration Driver --
                                                                               |
              <- targetPath ----------------------------------------------------
```

The libStorage integration driver would create a directory to mount the volume once it was attached (if block storage) or mount the volume directly if it was shared storage. 

**The path to the mounted volume was returned to Docker. This point is key.**

### 0.10.x CSI NodePublishOperation
A CSI `NodePublishVolume` operation looked like this:

```
CSI Client -> NodePublishVolume[targetPath] -> RR CSI Endpoint -> libStorage Bridge -> libStorage Integration Driver
```

Ultimately a CSI `NodePublishVolume` sat **on top** of our standard workflow. @codenrhoden  wrote very elegant code to then bind mount the mounted volume to the requested `targetPath`.

### 0.11.x Docker Mount Operation
Here's where it gets tricky for existing libStorage platforms:

```
Docker Client -> Mount -> RR Docker CSI bridge -> NodePublishVolume[targetPath] -> CSI Endpoint -> libStorage Integration Driver --
                                                                                                                                  |
              <- targetPath -------------------------------------------------------------------------------------------------------
```

Because REX-Ray's Docker endpoint now talks to CSI instead of libStorage directly,  the RR Docker->CSI bridge is the CSI client, or a CO. This means to perform a  `NodePublishOperation` the Docker endpoint must explicitly provide a `targetPath`.

### The Bug
This one is on me. When writing the Docker->CSI bridge I knew I needed to generate  a `targetPath`. At first I thought I needed to locate it outside the normal mount area defined by `libstorage.integration.volume.operations.mount.path`. Then I thought, "No, I'm overthinking it." Well, I wasn't.

The libStorage->CSI Bridge is bind mounting the private mount path to the supplied `targetPath`. And this is happening:

```shell
$ mount -o bind /var/lib/rexray/volumes/MyVolume/data /var/lib/rexray/volumes/MyVolume
```

In the end, the mount table will look like this:

```shell
SOURCE                                 MOUNT POINT
/dev/xvda1                             /var/lib/rexray/volumes/MyVolume
/var/lib/rexray/volumes/MyVolume/data  /var/lib/rexray/volumes/MyVolume
```

Here's the thing, the above technically works. You **can** bind mount a path back over itself. Except it screws up idempotency. The first unmount succeeds,  but successive, idempotent unmount attempts fail because the `targetPath` no longer exists since the first unmount removed the `targetPath` due to ref counting. And the idempotency provider for the libStorage->CSI bridge expects the `targetPath` to exist or an error is returned:

```go
var errMissingTargetPath  = errors.New("target path not created")

// IsNodePublished should return a flag indicating whether or
// not the volume exists and is published on the current host.
func (d *driver) IsNodePublished(
	ctx xctx.Context,
	id *csi.VolumeID,
	pubInfo *csi.PublishVolumeInfo,
	targetPath string) (bool, error) {

	_, err := os.Stat(targetPath)
	if os.IsNotExist(err) {
		return false, errMissingTargetPath
	}
```

## The Solution
Here's where I'm at with this:

1. I don't think the libStorage-CSI bridge's idempotency provider should necessarily return an error if the provided `targetPath` does not exist. I need to confirm this with @codenrhoden, but ultimately shouldn't this just mean the volume is not published on the node, and we should return false without an error?

2. To maintain backwards compatibility with environments that have already deployed REX-Ray we cannot change where REX-Ray's Docker integration driver mounts volumes. However, I suggest we introduce a new property, `rexray.csi.mount.path`, with a default value of `/var/lib/rexray/csi/volumes`, and use it as the root location for volumes mounted via REX-Ray's libStorage->CSI bridge. This means that if Docker wants to mount a volume named `MyVolume`, the system's mount table would look like this:

```shell
SOURCE                                 MOUNT POINT
/dev/xvda1                             /var/lib/rexray/volumes/MyVolume
/var/lib/rexray/volumes/MyVolume/data  /var/lib/rexray/csi/volumes/MyVolume
```

So going forward, new volumes created by the Docker->CSI bridge would be presented to Docker with a mount path rooted at  `/var/lib/rexray/csi/volumes`. However, the above change would also be backwards compatible with existing deployments.

Please provide your thoughts and comments below.

codenrhoden:
Thanks for the write-up @akutz.  It really helped me understand.

> I don't think the libStorage-CSI bridge's idempotency provider should necessarily return an error if the provided targetPath does not exist. I need to confirm this with @codenrhoden, but ultimately shouldn't this just mean the volume is not published on the node, and we should return false without an error?

It was considered an error because the CO shouldn't ever be sending us a targetPath that does not exist. However, I see no problem with the idempotency provider punting in this scenario and allowing the storage plugin to determine what to do in that scenario. If it's a publish request and the path doesn't exist, it is an error. If it's an unpublish request and the path doesn't exist, it's clearly not published there so the storage plugin can also check if it needs to remove it's private mount, etc...  So yeah, your proposal sounds fine.

> I suggest we introduce a new property, rexray.csi.mount.path, with a default value of /var/lib/rexray/csi/volumes

This makes sense to me. We will need to update all the docker plugins to set this as a propogated mount path as well. Seems like a clean solution to me. I didn't see this problem coming, that's for sure. In a way, we would now be using `/var/lib/rexray/volumes` as our private mount area, which is very consistent with what we've done in other drivers. 👍

clintkitson:
Re #2 these plugins would also need an updated propagated path to Docker.

clintkitson:
I would also be concerned with the root path parameter and ensure it still targeted `/var/lib/rexray/volumes/MyVolume/data` and `/data`.

akutz:
I am unclear on what both of you are referring to regarding propagating a path to Docker. The path is returned by our Docker endpoint. It's fine for Docker to consume volumes that exist in different areas on the filesystem. Old Docker-created volumes will be from our private area. New volumes will be created in the new area. Not sure what you two are seeing here...

clintkitson:
@akutz https://github.com/codedellemc/rexray/blob/master/.docker/plugins/cinder/config.json#L181

akutz:
Hi @clintkitson,

Thank you for the clarification. FWIW, your comment about `/data` is inconsequential because a volume mounted at `/var/lib/rexray/csi/volumes/VOLUME_NAME` would actually be bind mounted to `/var/lib/rexray/volumes/VOLUME_NAME/data`.

clintkitson:
Totally get that, but just making sure that the `/data` is still configurable after change, sounds like you have that covered.

akutz:
Hi @codenrhoden,

> It was considered an error because the CO shouldn't ever be sending us a targetPath that does not exist

That makes sense. I only considered the function in regards to Unpub. So yeah, we can make that distinction. Good call.

akutz:
Hi @clintkitson,

> Totally get that, but just making sure that the /data is still configurable after change, sounds like you have that covered.

Covered isn't fair since I'm kind of not doing anything. Just an extra bind mount. :) But yeah, it will work.

akutz:
Hi @codenrhoden,

> In a way, we would now be using /var/lib/rexray/volumes as our private mount area, which is very consistent with what we've done in other drivers.

That's *exactly* what we're going to be doing. Before we used it for both the private mount area **and** the Docker volume area. Except now since Docker is subject to the private mount area, they can no longer be the same.

codenrhoden:
@akutz I don't see anything in here that jumps out at me -- looks like you've covered all the bases, but only testing will find out!

PR #1056, opened 2017-10-09
Docker plugin updates

### LINKED ISSUE #933: docker: Add HTTP_PROXY to managed plug-ins
# Summary
As a golang binary, REX-Ray natively recognizes and uses the `HTTP_PROXY` env var. However, the Docker plugins only allow env vars that are explicitly defined in their config. `HTTP_PROXY` needs to be added to those configs so that users can pass it when their environment requires it.

### LINKED ISSUE #1043: docker: Enhance managed plug-in configuration options
The `config.json` file for each docker managed plugin lists all the allowed env vars for a plugin. Right now this list is restrictive, and is missing a lot of useful REX-Ray options, such as those for Linux file mode, options for data directory, etc.

We can either add more options to all plugins, but we can also explore auto-populating the vars from the output of `rexray env | grep <DRIVER>_`, plus some other standard ones.

- [ ] `HTTP_PROXY`

codenrhoden:
This patchset has a series of commits that update the Docker plugins. They are in separate commits for easier review, and potential to revert if something is wrong.

All Alpine-based plugins are upgraded from 3.5->3.6, *except* for s3fs, which does not compile on 3.6 due to the replacement of openssl with libressl.

The start scripts for REX-Ray no longer use the deprecated `-f` flag.

All drivers that use and HTTP API now support the `HTTP_PROXY` env var. This ends up being every plugin except Ceph RBD.

New global options for volume rootpath and filemode are added, except for S3FS which should be taken care of by #1042.

Many plugins had their options more fully brought in line with what is available to the full drivers -- but not all options apply to the Docker managed plugins, so there are still discrepancies.

Fixes #933 
Fixes #1043

### DIFF
--- a/.docker/plugins/Dockerfile
+++ b/.docker/plugins/Dockerfile
@@ -11,5 +11,5 @@ ADD rexray.yml /etc/rexray/rexray.yml
 ADD rexray.sh /rexray.sh
 RUN chmod +x /rexray.sh
 
-CMD [ "rexray", "start", "-f", "--nopid" ]
+CMD [ "rexray", "start", "--nopid" ]
 ENTRYPOINT [ "/rexray.sh" ]
--- a/.docker/plugins/Dockerfile
+++ b/.docker/plugins/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.5
+FROM alpine:3.6
 
 RUN apk update
 RUN apk add xfsprogs e2fsprogs ca-certificates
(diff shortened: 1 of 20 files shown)


codenrhoden:
This has been rebased after #1042.  It makes a small change from what #1042, which is to set the rootpath vars to an empty string, so that they use the default values in REX-Ray, rather than just explicitly setting them to `/data`, which is already the default.

PR #1060, opened 2017-10-11
Always use LocalDevices to get attached dev token

codenrhoden:
Rather than caching the attachment token returned by VolumeAttach, which
gets lots when REX-Ray is restarted, query the sytem's LocalDevices via
the driver Executor in the scenario where we know the volume is already
attached, because the device map returned always contains the token.

### DIFF
--- a/agent/csi/libstorage/csi_service_libstorage_idemp.go
+++ b/agent/csi/libstorage/csi_service_libstorage_idemp.go
@@ -17,6 +17,7 @@ var (
 	errMissingIDKeyPath   = errors.New("missing id key path")
 	errMissingTokenKey    = errors.New("missing token key")
 	errUnableToGetLocDevs = errors.New("unable to get local devices")
+	errLocDevsMissingVol  = errors.New("unable to find attached vol in local devices")
 	errMissingTargetPath  = errors.New("target path not created")
 )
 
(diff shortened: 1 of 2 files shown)


agent/csi/libstorage/csi_service_libstorage_idemp.go
```diff
@@ -118,26 +119,35 @@ func (d *driver) IsControllerPublished(
...
+		Opts:     apiutils.NewStore(),
+		ScanType: apitypes.DeviceScanQuick,
+	}
+	devs, err := d.client.Executor().LocalDevices(d.ctx, ldOpts)
+	if err != nil {
+		return nil, errUnableToGetLocDevs
 	}
+
+	for k, v := range devs.DeviceMap {
+		if v == vol.Attachments[0].DeviceName {
```
clintkitson: Should we range the attachments here too? ie do we care that the bridge may not do multi-attach?

akutz: Hi @clintkitson,

I see your point, except we've already declared that a COs cannot access a single Node as both CSI *and* Docker Volume API. It's one or the other since Docker's model is sufficiently different.

clintkitson: Thanks

akutz: FWIW, I've no problem adding a `// TODO` here and attempting to come back later to make the Docker bridge able to tolerate multiple mount points, but at the moment this is just one of a few locations where single-attachment is explicitly the case.

akutz: > a few locations

In the Docker bridge that is.

akutz: Hi @clintkitson,

One last thing, this is data derived from libStorage (its in the idemp provider for the LS bridge). Thus the LS driver would also have to provide support for multiple attachments. Most of ours do not.

PR #1063, opened 2017-10-12
nfs: Enable CSI-NFS for Docker

akutz:
This patch enables support for CSI-NFS for Docker.

### DIFF
--- a/.docker/plugins/csi-nfs/config.json
+++ b/.docker/plugins/csi-nfs/config.json
@@ -1,0 +1,120 @@
+{
+      "Args": {
+        "Description": "",
+        "Name": "",
+        "Settable": null,
+        "Value": null
+      },
+      "Description": "REX-Ray for CSI-NFS",
+      "Documentation": "https://github.com/codedellemc/rexray/.docker/plugins/csi-nfs",
+      "Entrypoint": [
+        "/rexray.sh", "rexray", "start", "--nopid"
+      ],
+      "Env": [
+        {
+          "Description": "",
+          "Name": "REXRAY_LOGLEVEL",
+          "Settable": [
+            "value"
+          ],
+          "Value": "warn"
+        },
+        {
+          "Description": "",
+          "Name": "LIBSTORAGE_SERVER_TASKS_EXETIMEOUT",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "",
+          "Name": "LIBSTORAGE_INTEGRATION_VOLUME_OPERATIONS_MOUNT_ROOTPATH",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "",
+          "Name": "LINUX_VOLUME_ROOTPATH",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "",
+          "Name": "LINUX_VOLUME_FILEMODE",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "",
+          "Name": "HTTP_PROXY",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
+          "Description": "The sock file used by the CO to communicate with CSI.",
+          "Name": "CSI_ENDPOINT",
+          "Settable": [
+            "value"
+          ],
+          "Value": ""
+        },
+        {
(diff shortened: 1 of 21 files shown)


.docker/plugins/csi-nfs/config.json
```diff
@@ -0,0 +1,136 @@
...
+          ],
+          "Value": "false"
+        },
+        {
+          "Description": "",
+          "Name": "LIBSTORAGE_SERVER_TASKS_EXETIMEOUT",
+          "Settable": [
+            "value"
+          ],
+          "Value": "120s"
```
codenrhoden: This shouldn't be needed.  The 120s is coming from Azure, where we needed to bump this beyond the default 60s.  I think leave the value empty so it stays at the default.

agent/csi/mod_csi_docker.go
```diff
@@ -386,14 +594,40 @@ func (d *dockerBridge) List() (*dvol.ListResponse, error) {
...
 
+		if _, ok := volMap[name]; ok {
+			d.ctx.WithField("volume", vi).Warn(
+				"docker-csi-bridge: List: skipped volume already in cache")
+			continue
+		}
+
+		d.ctx.WithField("volume", vi.Id.Values).WithField(
+			"name", name).Debug(
+			"docker-csi-bridge: List: found new volume")
```
codenrhoden: Do want to log every volume at this point?

codenrhoden: I had it it here before, but it was "new that wasn't in the cache", but that is completely wrong at this point.

agent/csi/mod_csi_docker.go
```diff
@@ -10,16 +11,47 @@ import (
...
 	"github.com/codedellemc/rexray/agent"
 	"github.com/codedellemc/rexray/core"
 	apictx "github.com/codedellemc/rexray/libstorage/api/context"
 	"github.com/codedellemc/rexray/libstorage/api/registry"
 	apitypes "github.com/codedellemc/rexray/libstorage/api/types"
+	dvol "github.com/docker/go-plugins-helpers/volume"
+)
+
+const (
+	rrCSIMountPath  = "rexray.csi.mount.path"
```
codenrhoden: This is unused at this point.

agent/csi/mod_csi_docker.go
```diff
@@ -10,16 +11,47 @@ import (
...
+	rrCSINFSVolumes = "rexray.csi.nfs.volumes"
+)
+
+func init() {
```
codenrhoden: This is duplicated with another `init` below. I can't believe Hound or any linter doesn't flag this. I hit the same thing in my rebase.

agent/csi/mod_csi_docker.go
```diff
@@ -386,25 +579,50 @@ func (d *dockerBridge) List() (*dvol.ListResponse, error) {
 			continue
 		}
 
-		vol := &dvol.Volume{Name: name}
+		if _, ok := volMap[name]; ok {
```
codenrhoden: Looks like this came back in. Sneaky code...  :)

agent/csi/mod_csi_docker.go
akutz: Hi @codenrhoden,

Well, at debug level it could be useful. What you're really asking, and I agree, is if there is some way we can finally introduce logging domains to RR. Yes. I mean, "Yes, we have to figure out how to do that."

codenrhoden: I still think this bit can go.

agent/csi/mod_csi_docker.go
```diff
@@ -386,25 +579,50 @@ func (d *dockerBridge) List() (*dvol.ListResponse, error) {
...
+	}
+
+	return d.buildListResponse(volMap)
+}
+
+func (d *dockerBridge) buildListResponse(
+	volMap map[string]csi.VolumeInfo) (*dvol.ListResponse, error) {
+
+	res := &dvol.ListResponse{}
+	res.Volumes = make([]*dvol.Volume, len(volMap))
```
codenrhoden: huzzah! 👍

agent/csi/mod_csi_docker.go
akutz: Did you remove it? If you did then I'm concerned that the patch didn't catch that. If everything else looks fine I'll chalk it up to my error.

akutz: Wait...

codenrhoden:
Build and Test in Travis-CI look good.

PR #1067, opened 2017-10-13
docker: enable CSI-NFS for docker

codenrhoden:
This replaces #1063

### DIFF
--- a/agent/csi/mod_csi_docker.go
+++ b/agent/csi/mod_csi_docker.go
@@ -442,6 +442,23 @@ func (d *dockerBridge) Create(req *dvol.CreateRequest) error {
 			return err
 		}
 
+		d.byNameRWL.Lock()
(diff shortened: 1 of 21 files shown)


agent/csi/mod_csi_docker.go
```diff
@@ -386,13 +579,30 @@ func (d *dockerBridge) List() (*dvol.ListResponse, error) {
...
 
+		if _, ok := volMap[name]; ok {
+			d.ctx.WithField("volume", vi).Warn(
+				"docker-csi-bridge: List: skipped volume already in cache")
+			continue
+		}
+
+		d.ctx.WithField("volume", vi.Id.Values).WithField(
+			"name", name).Debug(
+			"docker-csi-bridge: List: found new volume")
```
codenrhoden: I think the code from here up to the check if the name is in the map can be completely removed now.

PR #1076, opened 2017-10-16
CSI-VFS Managed Docker Plug-in

akutz:
This patch adds a Docker managed plug-in for CSI-VFS.


hpbieker:
Any idea when this will be merged?

.docs/user-guide/schedulers/docker/plug-ins/local.md
```diff
@@ -0,0 +1,59 @@
+# Local
+
+CSI-NFS, CSI-VFS
+
+---
+
+<a name="csi-nfs"></a>
```
hpbieker: Fix this link?

clintkitson:
Hello @hpbieker,

What was your intended use of the VFS plugin? For the most part we are using it for comprehensive end-to-end testing of the CSI plugins without a dependency on a backend storage platform. Any more info here would be great to here.

hpbieker:
@clintkitson. I am looking for a way to utilize a cifs network file system for my Docker volumes. I know I can use the local volume driver and mount the cifs file system in each container. However, this means that I will have to include the username and password in the docker-stack.yml file for each service. I was hoping to just define a volume driver that could have predefined the authentication to my cifs file store.

I guess this would be possible by using the vfs driver. Preferably, there is a cifs driver similar to the nfs driver so that the mounting of the cifs drive is handled automatically. I guess it is just a matter of generalization of the nfs driver.

clintkitson:
What does "predefined the authentication to my cifs file store" mean?

hpbieker:
@clintkitson Sorry for the bad wording. I mean that I do not want to specify a username, password, hostname and path each volume I create. I think this should be a part of the configuration of the volume driver.

PR #1083, opened 2017-10-24
Fix RBD CephArgs string test for userid

mvollman:
The strContainsClient function is spliting characters instead of words
and not correctly determining if a user has been defined.

PR #1098, opened 2017-10-27
ceph: Fix how ceph CLI args are parsed

mvollman:
The `strContainsClient` function is splitting characters instead of words and not correctly determining if a user has been defined.

I mistakenly closed the original merge request #1083 while rebasing.

### DIFF
--- a/libstorage/drivers/storage/rbd/utils/utils_test.go
+++ b/libstorage/drivers/storage/rbd/utils/utils_test.go
@@ -75,3 +75,47 @@ func TestParseMonitorAddresses(t *testing.T) {
 		assert.True(t, test.ip.Equal(ip[0]))
 	}
 }
+
+func TestArgsClient(t *testing.T) {
+	tests := []struct {
+		args    string
+		present bool
+	}{
+		{
+			args:    "--id myuser",
+			present: true,
+		},
+		{
+			args:    "--user myuser",
+			present: true,
+		},
+		{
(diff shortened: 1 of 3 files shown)


akutz:
Hi @mvollman,

Unless the commits need to remain separate, do you mind please squashing them? Thanks!

libstorage/drivers/storage/rbd/utils/utils_test.go
```diff
@@ -75,3 +75,43 @@ func TestParseMonitorAddresses(t *testing.T) {
...
+			args:    "--user myuser",
+			present: true,
+		},
+		{
+			args:    "--name client.myuser",
+			present: true,
+		},
+		{
+			args:    "--cluster cluster2",
+			present: true,
```
codenrhoden: This should be `present: false`. There is only a `cluster` flag in here, and therefore no client is detected. This should actually fail if you run the tests, e.g. :
`go test -v github.com/thecodeteam/rexray/libstorage/drivers/storage/rbd/utils`

Do you not see the tests fail?

I also had a 6th variation in the sample I pasted:
```
		{
			args:    "--cluster cluster2 --id ceph",
			present: true,
		},
```

That was just to make sure we catch a mixed example.

mvollman:
@codenrhoden What is the one outstanding request?

codenrhoden:
@mvollman Nothing! You took care of it in your most recent commit.  Thank you!

PR #1110, opened 2017-11-10
Fix disablement of path caching

codenrhoden:
Even though there was a config option to disable path caching, it was
essentially being ignored. There were competing scenarios that forced
path caching to always be enabled no matter.

This patch achieves the desired scenario of:
- If running a one-off CLI command (e.g. "rexray volume ...") path caching
is disabled entirely.
- If REX-Ray is run as a service, path caching is controlled via config
  options, and still currently defaults to "enabled".

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -33,7 +33,6 @@ func Start(
 
 	// Enable path caching for the modules.
 	config = config.Scope(configRexrayAgent)
-	config.Set(apitypes.ConfigIgVolOpsPathCacheEnabled, true)
 
 	// Activate libStorage if necessary.
 	var (
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -289,6 +289,10 @@ func (c *CLI) updateLogLevel() {
 }
 
 func (c *CLI) preRunActivateLibStorage(cmd *cobra.Command, args []string) {
+	// Disable patch caching for the CLI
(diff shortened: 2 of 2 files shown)


codenrhoden:
Even when disabling path caching via config, it was still enabled because of https://github.com/thecodeteam/rexray/blob/bcae8bae40165981c961006f198585b54a15d248/agent/agent.go#L36

however, that was in direct conflict with https://github.com/thecodeteam/rexray/blob/master/cli/cli.go#L305

By moving the code to explicitly disable path caching from `preRun` to `preRunActivateLibStorage`, we only explicitly disable path caching when calling the adapter, device, volume, and snapshot subcommands. For all others, we defer to the config.

This may have the nice side effect of speeding up `volume ls` considerably, since it was essentially happening twice!

cduchesne:
The change worked as expected. Only feedback I have is I believe we should make path cache disabled by default. If this is something we can consider, I will test different version of docker (older) to compare how it works then vs now.

codenrhoden:
@cduchesne I'd rather not change the default behavior of REX-Ray in a patch release, but your suggestion may be a sound one. Let's file an issue for it and revisit for an upcoming release.

@akutz any objections to getting this one merged?

codenrhoden:
@akutz Rebased this one, so it should be in the clear. See the original description and very first comment for details on what was happening here.

PR #1111, opened 2017-11-13
Use correct uname machine hardware name for s390x

michael-holzheu:
Currently the string "S390X" is used which leads to the following build error:

```
# make rpm
setarch: S390X: Unrecognized architecture
Makefile:123: recipe for target 'rexray-0.11.0+13-1.S390X.rpm' failed
```

The correct uname machine name is "s390x":

```
# uname -m
s390x
```

So fix this an use the correct name.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -166,5 +166,5 @@ var goarchToUname = map[string]string{
 	"mipsle":   "MIPS32LE",
 	"ppc64":    "PPC64",
 	"ppc64le":  "PPC64LE",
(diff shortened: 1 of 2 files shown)


michael-holzheu:
Any news on that issue?

clintkitson:
@michael-holzheu We are looking to do a 0.11.1 shortly. @codenrhoden what do you think, is this something that might be able to be targeted? Do we know what the scope of this change might be?

PR #1124, opened 2017-12-08
Add more spacing for a better readability

jmaitrehenry:
After reading the documentation, I found the information hard to find, but not because the search engine or the structure don't work, but because the text was hard to read.

I adjust some CSS rules:
- adding more spacing and for having a font-size a little bigger
- adjust the size for having the same inside and outside tables
- simplify `th` text alignment*
- rework the hX element for having a better distinction between elements and for having an h4 more visible

*I see some places where we use `<center>title</center>` syntax for th cell. I remove the rules inside `bootstrap-custom.min.css` for force align the text to the right.

Some before/after screenshots:
screen shot 2017-12-08 at 2 09 52 pm
screen shot 2017-12-08 at 2 10 10 pm
screen shot 2017-12-08 at 2 10 24 pm
screen shot 2017-12-08 at 2 10 50 pm


akutz:
Hi @jmaitrehenry,

Please rebase this onto `master`. After that, I will need to ponder this a bit. I agree with some of your changes, while others I find a matter of preference.

jmaitrehenry:
@akutz Done, let's discuss on the changes, maybe I can split this PR in two, the first one on thing we agree, and the other one as a proposal where we can discuss.

akutz:
Hi @jmaitrehenry,

Instead of modifying a minified CSS file I'd prefer you override those rules with `!important` in a separate file. Or would that not work?

akutz:
Hi @jmaitrehenry,

FYI, I've no issues with this PR other than the one I raised above (the modification to a minified CSS file). Once we resolve that discussion I will merge the change. Thank you again for this!

jmaitrehenry:
@akutz great, I try to find the change on the minified css, I don't know why is it modified, I do not do that normally.

akutz:
Hi @jmaitrehenry,

Actually, I take all that back :) Will you please first test the changes by enabling your browser's dev mode and verifying the updates don't break mobile mode? For example, here's Safari's *Responsive Design Mode*:

image

Chrome also has this ability.

akutz:
Hi @jmaitrehenry,

> I try to find the change on the minified css, I don't know why is it modified, I do not do that normally.

You mean you didn't intend to modify that file? Except you say here that you **did** modify it...

> I see some places where we use <center>title</center> syntax for th cell. I remove the rules inside bootstrap-custom.min.css for force align the text to the right.

jmaitrehenry:
Oh didn't see it, I made it months ago. Ok, let me check on mobile and remove the modification inside .min file.

akutz:
Hi @jmaitrehenry,

Thank you. Please see if the change can be made using `!important` in one of the custom CSS files. If not...I mean...hmm. I just hate modifying stock files like that. Especially when they're minified. It makes it difficult to ensure changes are persisted if those stock files are updated.

jmaitrehenry:
For mobile:
screen shot 2018-06-22 at 12 04 11 pm
screen shot 2018-06-22 at 12 04 25 pm
screen shot 2018-06-22 at 12 04 57 pm
screen shot 2018-06-22 at 12 05 15 pm
screen shot 2018-06-22 at 12 05 34 pm
screen shot 2018-06-22 at 12 05 51 pm
screen shot 2018-06-22 at 12 06 04 pm

I will change the table rendering for phone, it's too big and the table is less readable than before.

PR #1133, opened 2017-12-19
Use CentOS as base for csi-nfs docker plugin

### LINKED ISSUE #1114: How to use rexray/csi-nfs with docker plugin ?
I'm tyring to use rexray for my docker and NFS server.
I found this page, https://rexray.readthedocs.io/en/stable/user-guide/schedulers/docker/plug-ins/csi-nfs/ .
But it is not clear to me : (
Anyone can inform me how to use it by docker plugin ?

codenrhoden:
Add docs that show examples of using the plugin.

Fixes #1114

### DIFF
--- a/.docs/user-guide/schedulers/docker/plug-ins/csi-nfs.md
+++ b/.docs/user-guide/schedulers/docker/plug-ins/csi-nfs.md
@@ -27,3 +27,54 @@ The following environment variables can be used to configure the plug-in:
 | Environment Variable | Description | Default | Required |
 |---------------------|-------------|---------|---------|
 | `X_CSI_NFS_VOLUMES` | A list of NFS volume mappings | | |
+
+## Examples
+
+To use the `csi-nfs` plug-in, you must first create a volume using the
+`docker volume create` command (unless predefined using `X_CSI_NFS_VOLUMES`, see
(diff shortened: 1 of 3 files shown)


.docs/user-guide/schedulers/docker/plug-ins/csi-nfs.md
```diff
@@ -27,3 +27,54 @@ The following environment variables can be used to configure the plug-in:
...
+```
+
+### Limitations
+
+`csi-nfs` is an early proof of concept to demonstrate a Docker managed plug-in functioning with [CSI](https://github.com/container-storage-interface/spec). It is
+not intended to replace Docker's
+[native NFS](https://docs.docker.com/engine/reference/commandline/volume_create/#driver-specific-options)
+functionality.
+
+Currently, the plug-in only supports NFSv4, and does not allow for custom options
```
briantopping: I think one of the unique opportunities that CSI-NFS has is with non-Swarm orchestration environments. Maintaining what could amount to hundreds of volumes with `docker volume create` commands would be onerous, less so in a RR master config (if I understand the configuration correctly that there can be a central RR that others proxy to). Or am I misapplying RR in that it's really only intended as a multiplexer to reduce driver development tasks?

Issue #1118, opened 2017-11-30
rexray/ebs plugin does not start if HTTP_PROXY value is set


spatialvlad:
# Summary
We have several Docker Swarms for our US Prod and Dev environments. Each of the swarms has several volumes created with the help of rexray/ebs docker volume plugin. Everything works fine, volumes are created, automatically re-mapped, etc.
Recently I started creating a Docker Swarm for our China AWS environment. However, I cannot make rexray/ebs plugin work for the China AWS. 

Without modifying **HTTP_PROXY** variable the plugin tries to connect to https://ec2.cn-north-1.amazonaws.com AWS API, but it does not work, since the correct URL for the API is https://ec2.cn-north-1.amazonaws.com.cn

When I set **HTTP_PROXY** environment variable it is not possible to enable rexray/ebs. It is happens both when setting the HTTP_PROXY during plugin install and setting it later.
I tried all versions of the plugin that support setting HTTP_PROXY starting from 0.11.0-rc1 and up to edge.
I also tried using both http and https, as well as any url for it, e.g. https://ec2.us-west-2.amazonaws.com
I also tried rebuilding the instance with fresh reinstalls of Docker.
Whenever there is a value set for HTTP_PROXY, the plugin fails to be enabled. It starts to work again as soon as I empty HTTP_PROXY by setting HTTP_PROXY= and enabling it.

## Version
**Plugin:**
rexray/ebs versions 0.11.0-rc1, 0.11.0-rc2, 0.11.0-rc3, 0.11.0, latest, edge 

**OS:**
Ubuntu 16.04.3 LTS (GNU/Linux 4.4.0-101-generic x86_64)

**Docker version:**
```shell
Client:
 Version:      17.09.0-ce
 API version:  1.32
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:42:18 2017
 OS/Arch:      linux/amd64

Server:
 Version:      17.09.0-ce
 API version:  1.32 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   afdb6d4
 Built:        Tue Sep 26 22:40:56 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

## Expected Behavior
Plugin starts to use HTTP_PROXY url for AWS API calls when provided.

## Actual Behavior
Plugin fails to start.

## Steps To Reproduce
1. Install latest rexray/ebs docker driver plugin
[...]

PR #1134, opened 2017-12-19
Give EBS endpoint precedence over region

codenrhoden:
If a user has configured an endpoint to use for AWS, always use it,
rather than deferring to an auto-generated endpoint based on the region.


codenrhoden:
@akutz would like you to review this one. This makes it so that if a user sets `ebs.endpoint`, that endpoint is actually used, always. Previously, even if it was set, it would be ignored if there was a region set, *or* if there was a region in the context, which for REX-Ray is pretty much always true since it is part of the IID metadata.

I contend that if a user sets the endpoint, we should use that endpoint no matter way.

codenrhoden:
This should fix #1118 (really really this time)

PR #1168, opened 2018-02-06
Add support to build the RBD plugin on AArch64

arm64b:
Update the Dockerfile in favor of the multi-arch images
to support build the RBD plugins on AArch64 platforms:

`$DRIVER=rbd make build-docker-plugin`
Have been tested on both amd64 and arm64.

Signed-off-by: Dennis Chen <dennis.chen@arm.com>

### DIFF
--- a/.docker/plugins/rbd/.Dockerfile
+++ b/.docker/plugins/rbd/.Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:7.4.1708
+FROM centos:latest
 
 ENV CEPH_VERSION luminous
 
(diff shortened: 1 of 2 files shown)


clintkitson:
Thanks for the PR @arm64b!

.docker/plugins/rbd/.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:7.4.1708
+FROM centos:latest
```
akutz: Hi @arm64b,

I'm not sure we want to pull the latest. Why make this change? I prefer that `FROM` statements refer to a specific version for purposes of reproducibility. @codenrhoden?

arm64b: Ah, this should be stale. I think we can close this PR now 🛩

arm64b:
In a stale status, so close it now

PR #1170, opened 2018-02-09
Skip unauthorized pools in the rbd driver.

highfly22:
Try to use rexray in the non-privilege account.


akutz:
Hi @highfly22,

I like the change, but I wonder if @codenrhoden thinks it should be configurable? If so, default to skip/error?

Please go ahead and rebase onto `master` so the PR is up-to-date prior to merge. Thanks!

codenrhoden:
@highfly22 I would prefer to see this as configurable. If you just blindly ignore the inability to list images in a pool, you have no way to know if it's because your Auth key doesn't allow you access to that specific pool, or if your key is completely invalid.

A config option to enable skipping inaccessible pools would be good, and I would want to see a debug log entry saying that the pool was skipped, such that if you turn on debugging you see it happening.

akutz:
@highfly22 ping

PR #1199, opened 2018-03-30
Improve initscript behavior

jggc:
Current initscript does not work as intended :

 - When is starts the service it stays in the foreground, initscripts
   should run daemons in the background
 - It logs to the tty, initscripts should redirect their output to
   somewhere that makes sense.

So here is this patch that will send the appropriate targets of the
rexray service in the background and only output minimal informative
logging to the tty, any other logs are redirected into
/var/log/rexray/initscript.log

Also refactored to use "${}" across the script

### DIFF
--- a/cli/installer.go
+++ b/cli/installer.go
@@ -398,7 +398,8 @@ func createInitFile() error {
 	return nil
 }
 
-const initScriptTemplate = `### BEGIN INIT INFO
+const initScriptTemplate = `#!/bin/bash
+### BEGIN INIT INFO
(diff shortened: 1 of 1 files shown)


clintkitson:
Thank you!

akutz:
Hi @jggc,

I'm happy to accept this PR, but please:

1. Rebase onto `master`
2. Split the current commit into two:
    * Commit 1 - Amazon Linux Support
    * Commit 2 - The global change to `${}` 

Thank you!

akutz:
Ping @jggc

PR #1203, opened 2018-04-04
Update aws.md

alanhughes:
It was not possible to run rexray on 5th generation AWS instances, which our relevant infrastructure uses. An early warning in the documentation would have been super handy!


clintkitson:
Thanks for the note @alanhughes, we will make sure to get this in the docs.

PR #1232, opened 2018-06-04
Fix JSON parser for rbd showmapped in Ceph v13

### LINKED ISSUE #1231: Rexray fails to parse output from Ceph 13 (Mimic) release
# Summary
Rexray fails to parse output from Ceph 13 (Mimic) release.

# New Feature
N/A

# Bug Reports
The JSON format of `rbd showmapped` changed in Ceph 13 (see http://docs.ceph.com/docs/master/releases/mimic/ and https://github.com/ceph/ceph/pull/19937). The old format was:
```json
{"0":{"pool":"luminpool","name":"fooimg","snap":"-","device":"/dev/rbd0"},"1":{"pool":"luminpool","name":"barimg","snap":"-","device":"/dev/rbd1"}}
```
and the new format is (for essentially the same setup):
```json
[{"id":"0","pool":"mimicpool","name":"fooimg","snap":"-","device":"/dev/rbd0"},{"id":"1","pool":"mimicpool","name":"barimg","snap":"-","device":"/dev/rbd1"}]
```
The JSON unmarshaller in rexray/libstorage/drivers/storage/rbd/utils/utils.go:166 cannot handle the new format.

## Version
```shell
rexray version
INFO[0000] updated log level                             logLevel=debug
DEBU[0000] os.args                                       time=1527869986744 val=[rexray version]
REX-Ray
-------
Binary: /usr/bin/rexray
SemVer: 0.11.2
OsArch: Linux-x86_64
Commit: fc8bfbd2d02c2690fc3a755a9560dd12c88e0852
Formed: Sat, 24 Feb 2018 23:51:22 UTC
DEBU[0000] completed cli execution                       time=1527869986744
INFO[0000] exiting process                               time=1527869986744
DEBU[0000] completed onExit at end of program            time=1527869986744
```

## Expected Behavior
Commands like `rexray volume ls` will return a list of mapped RBD images.

## Actual Behavior
`rexray volume ls` returns an error instead:
`FATA[0001] unable to parse rbd showmapped                inner=json: cannot unmarshal array into Go value of type map[string]*utils.rbdMappedEntry`

## Steps To Reproduce
Please list the steps to reproduce the issue in this section.

1. Set up Ceph 13 and map a RBD image
2. Set the Rexray config to detect RBD volumes
3. Try to run `rexray volume ls` and it should fail with a parsing error

## Configuration Files
```yaml
rexray:
  loglevel: debug
  service: rbd
rbd:
  defaultPool: mimicpool
```

### Service Log
(running in foreground mode)
https://gist.github.com/templed/427a1d55e414da4365a54a3aaf0d67b2

### Client Log
https://gist.github.com/templed/0047bfc2902acb7ba284b565b87f6cd9

templed:
Fix #1231 

I am not very experienced in Golang so please let me know if there are problems (particularly the return type of `ParseMappedRBDs` is something that worries me).

### DIFF
--- a/libstorage/drivers/storage/rbd/utils/utils.go
+++ b/libstorage/drivers/storage/rbd/utils/utils.go
@@ -36,7 +36,15 @@ var (
 	ctxConfigKey = interface{}("rbd.config")
 )
 
+type preMimicRbdMappedEntry struct {
+	Device string `json:"device"`
+	Name   string `json:"name"`
+	Pool   string `json:"pool"`
+	Snap   string `json:"snap"`
+}
+
 type rbdMappedEntry struct {
+	ID     string `json:"id"`
 	Device string `json:"device"`
 	Name   string `json:"name"`
 	Pool   string `json:"pool"`
@@ -161,12 +169,10 @@ func GetMappedRBDs(ctx types.Context) (map[string]string, error) {
 	}
 
 	devMap := map[string]string{}
-	rbdMap := map[string]*rbdMappedEntry{}
+	rbdMap, parseErr := parseMappedRBDs(out)
 
-	err = json.Unmarshal(out, &rbdMap)
-	if err != nil {
-		return nil, goof.WithError(
-			"unable to parse rbd showmapped", err)
+	if parseErr != nil {
+		return nil, parseErr
 	}
 
 	for _, mapped := range rbdMap {
@@ -175,6 +181,33 @@ func GetMappedRBDs(ctx types.Context) (map[string]string, error) {
 	}
 
 	return devMap, nil
+}
+
+//parseMappedRBDs is a two-stage parse because Ceph v13 changed JSON format
+func parseMappedRBDs(output []byte) ([]*rbdMappedEntry, error) {
+	rbdMap := []*rbdMappedEntry{}
+	rbdMapPreMimic := map[string]*preMimicRbdMappedEntry{}
+
+	err := json.Unmarshal(output, &rbdMapPreMimic)
+	if err != nil {
+
+		err2 := json.Unmarshal(output, &rbdMap)
+		if err2 != nil {
+			return nil, goof.WithError(
+				"unable to parse rbd showmapped", err2)
+		}
+
+	} else {
+		for index, entry := range rbdMapPreMimic {
+			rbdMap = append(rbdMap, &rbdMappedEntry{
+				ID: index, 
+				Device: entry.Device, 
+				Name: entry.Name, 
+				Pool: entry.Pool, 
+				Snap: entry.Snap})
+		}
+	}
+	return rbdMap, nil
 }
 
 //RBDCreate creates a new RBD volume on the cluster
--- a/libstorage/drivers/storage/rbd/utils/utils_test.go
+++ b/libstorage/drivers/storage/rbd/utils/utils_test.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"net"
+	"reflect"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


libstorage/drivers/storage/rbd/utils/utils.go
```diff
@@ -177,6 +175,34 @@ func GetMappedRBDs(ctx types.Context) (map[string]string, error) {
 	return devMap, nil
 }
 
+//ParseMappedRBDs is a two-stage parse attempt because Ceph v13 has different JSON format to previous versions
+func ParseMappedRBDs(output *[]byte) (map[string]*rbdMappedEntry, error) {
```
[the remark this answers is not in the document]
codenrhoden: This warning is coming because `ParseMappedRBDs` is capitalized. That makes it an exported func, and it doesn't need to be. Lower case the first letter and this will be fine.

I would also change it such that your input var, output, should be of type `[]byte`. There is no reason to use a pointer to a slice in Go, as a slice is already essentially a pointer.

libstorage/drivers/storage/rbd/utils/utils.go
```diff
@@ -177,6 +175,34 @@ func GetMappedRBDs(ctx types.Context) (map[string]string, error) {
...
+	rbdMap := map[string]*rbdMappedEntry{}
+
+	err := json.Unmarshal(*output, &rbdMap)
+	if err != nil {
+		type v13RbdMappedEntry struct {
```
codenrhoden: I'd rather see this type defined at the type, right after the origijal `rbdMappedEntry` struct.

I might suggest renaming the old struct to `rbdMappedEntryPreMimic` to denote it's the "old" version, but that's your call. no biggie on that one.

libstorage/drivers/storage/rbd/utils/utils.go
```diff
@@ -177,6 +175,34 @@ func GetMappedRBDs(ctx types.Context) (map[string]string, error) {
...
+		}
+		rbdMapv2 := []*v13RbdMappedEntry{}
+		err2 := json.Unmarshal(*output, &rbdMapv2)
+		if err2 != nil {
+			return nil, goof.WithError(
+				"unable to parse rbd showmapped", err2)
+		}
+		for _, entry := range rbdMapv2 {
+			tmp := rbdMappedEntry{Device: entry.Device, Name: entry.Name, Pool: entry.Pool, Snap: entry.Snap}
+			rbdMap[entry.ID] = &tmp
```
codenrhoden: the `tmp` var isn't really necessary here. Just do `rbdMap[entry.ID] = &rbdMappedEntry{...}`

templed:
I've made the changes requested and suggested. 

With regards to your suggestion in my original commit I was hesitant at changing the main rbdMappedEntry type as I was concerned it might be used elsewhere, hence me only having that type within the parser and transforming the new format into the older one. I've inverted that process (so now the old format is transformed into the new one) since I feel that makes more sense given the new format is, well, newer.

libstorage/drivers/storage/rbd/utils/utils.go
```diff
@@ -177,6 +183,28 @@ func GetMappedRBDs(ctx types.Context) (map[string]string, error) {
...
+
+		err2 := json.Unmarshal(output, &rbdMap)
+		if err2 != nil {
+			return nil, goof.WithError(
+				"unable to parse rbd showmapped", err2)
+		}
+
+	} else {
+		for index, entry := range rbdMapPreMimic {
+			rbdMap = append(rbdMap, &rbdMappedEntry{ID: index, Device: entry.Device, Name: entry.Name, Pool: entry.Pool, Snap: entry.Snap})
```
codenrhoden: Since you are going to need to squash and rebase anyways, I'd like to request a stylistic change. Please try to adhere to the 80 character max style, and put this line into multiple lines, akin to [this example](https://github.com/rexray/rexray/blob/6340189e4ae317293cb005eb7e30230380250d27/libstorage/drivers/storage/rbd/storage/rbd_storage.go#L143-L147).

meaning, like this:
```go
		        rbdMap = append(rbdMap, &rbdMappedEntry{
                             ID: index,
                             Device: entry.Device,
                             Name: entry.Name,
                             Pool: entry.Pool,
                             Snap: entry.Snap})
```

templed:
Hi @codenrhoden 

Everything should be good to go. Let me know if I did something silly and I'll fix it.

Thanks!

PR #1233, opened 2018-06-07
Support NVME EBS Volumes on M5 / C5

JustinVenus:
With custom udev rules[1] and nvme-cli, this change set provides support for NVME EBS volumes on C5/M5 families of EC2 instances.

#### Example udev rule
```sh
# /etc/udev/rules.d/999-aws-ebs-nvme.rules
# ebs nvme devices
KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", ATTRS{model}=="Amazon Elastic Block Store", PROGRAM="/usr/local/bin/ebs-nvme-mapping /dev/%k", SYMLINK+="%c"
```

#### Helper script for udev rules
```sh
#!/bin/bash
#/usr/local/bin/ebs-nvme-mapping
vol=$(/usr/sbin/nvme id-ctrl --raw-binary "$1" | cut -c3073-3104 | tr -s ' ' | sed 's/ $//g')
vol=${vol#/dev/}
if [[ -n "$vol" ]]; then
  echo ${vol/xvd/sd} ${vol/sd/xvd}
fi
```

#### Based on:
1) https://gist.github.com/jalaziz/bcfe2f71e3f7e8fe42a9c294c1e9279f

### DIFF
--- a/libstorage/drivers/storage/ebs/utils/utils_unix.go
+++ b/libstorage/drivers/storage/ebs/utils/utils_unix.go
@@ -4,8 +4,10 @@ package utils
 
 import (
 	"regexp"
+	"strings"
 
 	"github.com/rexray/rexray/libstorage/api/types"
+	log "github.com/sirupsen/logrus"
 )
 
 // DeviceRange holds slices for device namespace iteration
@@ -49,10 +51,27 @@ var (
 		},
 		DeviceRE: regexp.MustCompile(`^xvd[f-p]$`),
 	}
+	nvmeDeviceRange = &DeviceRange{
+		ParentLetters: []string{"0"},
+		ChildLetters: []string{
+			"0", "1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13",
+			"14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26"},
+		NextDeviceInfo: &types.NextDeviceInfo{
+			Prefix:  "nvme",
+			Pattern: "[0-26]n.*",
+			Ignore:  false,
+		},
+		DeviceRE: regexp.MustCompile(`^nvme[0-26]n.*`),
+	}
 )
 
 // GetDeviceRange returns a specified DeviceRange object
-func GetDeviceRange(useLargeDeviceRange bool) *DeviceRange {
+func GetDeviceRange(useLargeDeviceRange bool, instanceType string) *DeviceRange {
+	log.Debug("InstanceType: ", instanceType)
+	if strings.HasPrefix(instanceType, "c5") || strings.HasPrefix(instanceType, "m5") {
+		log.Debug("nvme device")
+		return nvmeDeviceRange
+	}
 	if useLargeDeviceRange {
 		return largeDeviceRange
 	}
--- a/libstorage/drivers/storage/ebs/executor/ebs_executor.go
+++ b/libstorage/drivers/storage/ebs/executor/ebs_executor.go
@@ -15,6 +15,7 @@ import (
 	"github.com/akutz/goof"
 	log "github.com/sirupsen/logrus"
 
+	"github.com/rexray/rexray/libstorage/api/context"
 	"github.com/rexray/rexray/libstorage/api/registry"
 	"github.com/rexray/rexray/libstorage/api/types"
 	"github.com/rexray/rexray/libstorage/drivers/storage/ebs"
@@ -49,7 +50,8 @@ func (d *driver) Init(ctx types.Context, config gofig.Config) error {
 	// initialize device range config
 	useLargeDeviceRange := d.config.GetBool(ebs.ConfigUseLargeDeviceRange)
 	log.Debug("executor using large device range: ", useLargeDeviceRange)
-	d.deviceRange = ebsUtils.GetDeviceRange(useLargeDeviceRange)
+	d.deviceRange = ebsUtils.GetDeviceRange(
+		useLargeDeviceRange, d.InstanceType(ctx))
 	return nil
 }
 
@@ -72,6 +74,15 @@ func (d *driver) InstanceID(
 	ctx types.Context,
 	opts types.Store) (*types.InstanceID, error) {
 	return ebsUtils.InstanceID(ctx, d.Name())
+}
+
+func (d *driver) InstanceType(ctx types.Context) string {
+	if iid, ok := context.InstanceID(ctx); ok {
+		if v, ok := iid.Fields["instanceType"]; ok && v != "" {
+			return v
+		}
+	}
+	return "default"
 }
 
 var errNoAvaiDevice = goof.New("no available device")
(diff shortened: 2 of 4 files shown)


JustinVenus:
The output from `continuous-integration/travis-ci/pr` appears flakey to me.

libstorage/drivers/storage/ebs/utils/utils_unix.go
```diff
@@ -49,10 +51,27 @@ var (
...
+		},
+		DeviceRE: regexp.MustCompile(`^nvme[0-26]n.*`),
+	}
 )
 
 // GetDeviceRange returns a specified DeviceRange object
-func GetDeviceRange(useLargeDeviceRange bool) *DeviceRange {
+func GetDeviceRange(useLargeDeviceRange bool, instanceType string) *DeviceRange {
+	log.Debug("InstanceType: ", instanceType)
+	if strings.HasPrefix(instanceType, "c5") || strings.HasPrefix(instanceType, "m5") {
```
jippi: i wonder if we could make this less static - i imagine m5 and other instances doing NVMe too going forward 

would it be sufficient to stat if `/dev/nvme0` exist ?

JustinVenus: I know i3's use nvme ephemeral drives, but I'm not certain how EBS might be presented to this family.  I'll investigate and follow up.

clintkitson:
Thank you for the contribution. Looking forward to agreement on the solution.

libstorage/drivers/storage/ebs/utils/utils_unix.go
jippi: this is from an `i3.2xlarge`

```
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev             30G   12K   30G   1% /dev
tmpfs           6.0G   75M  6.0G   2% /run
/dev/xvda1       99G   15G   80G  16% /
none            4.0K     0  4.0K   0% /sys/fs/cgroup
none            5.0M     0  5.0M   0% /run/lock
none             30G     0   30G   0% /run/shm
none            100M     0  100M   0% /run/user
/dev/nvme0n1    1.8T  1.6T  149G  92% /data
```

`/data` is the local instance data
`/dev/xvda1` is the EBS root volume

jippi: my main concern is that having rexray hardcode instance families for using nvme or not is kinda rough adoption wise, since rexray don't cut releases that often

codenrhoden: @jippi apologies that it has been so long, indeed. But we are ramping back up. =)

I need to come up to speed on how NVMe devices are working on AWS before providing a proper review. It does appear complicated if they are handled differently depending on the instance type...  :/

jippi: with the `udev` rule in OP, that is also bundled in AWS Linux handles mapping the nvme EBS device to the classic `/dev/*vda*` device - so if reading the type of the rexray expected `/dev/` device and checking if its a symlink to a nvme device - it should be generic enough to handle all instance types.

The reason why it does not work atm is that `/proc/partitions` show the nvme device, not the `virtual` device the `udev` rule creates - if its modified to `resolve` the non-nvme device name to the real nvme device name, it should just work (tm)

jippi:
@JustinVenus @clintkitson @codenrhoden Any progress/thoughts on this ? :)

JustinVenus:
Sorry, I've been super busy and forgot to follow up.  I'm not sure when I'm going to be able to circle back to this issue.  FWIW I've been running this patch in production w/o issue.

jippi:
@JustinVenus currently its broken because it doesn't whitelist r5 and z1 instances too

stsimb:
The travis-ci automated tests that failed are not actually related to the PR, it's transient errors pulling some debian packages from a mirror.

@JustinVenus maybe you could add r5 and z1 instances in the PR to trigger travis-ci running again.

jippi:
@JustinVenus i've tried your branch locally, with the identical config, i can't get it to work - you sure you don't have any addition config?

Crapworks:
We are currently hitting this as well, and it seems like the same problems exist for c4. Can you confirm that? Or are we hitting another problem?

jippi:
we use c4 with rexray/ebs just fine in prod

jippi:
For visibility i've created an alternative PR (#1252)

JustinVenus:
@jippi my configuration looks like

```txt
rexray:
  storageDrivers:
  - ec2
  modules:
    default-csi:
      host: unix:///run/rexray/rexray.sock
libstorage:
  service: ebs
  integration:
    volume:
      operations:
        mount:
          preempt: true
ebs:
  maxRetries:         10
  accessKey:          XXXXXXXXXXXXXXXXXXXX
  secretKey:          XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
  statusMaxAttempts:  10
  statusInitialDelay: 100ms
  statusTimeout:      7500ms
```

JustinVenus:
@jippi  Everything about one of my working hosts.
```sh
[fedora@mesos-master01fu2 ~]$ mount | grep rexray
/dev/nvme1n1 on /var/lib/rexray/volumes/mesos-master01fu2.aws.sig type ext4 (rw,relatime)

[fedora@mesos-master01fu2 ~]$ curl 169.254.169.254/latest/meta-data/instance-type
m5.xlarge

[fedora@mesos-master01fu2 ~]$ cat /etc/udev/rules.d/999-aws-ebs-nvme.rules
# ebs nvme devices
KERNEL=="nvme[0-9]*n[0-9]*", ENV{DEVTYPE}=="disk", ATTRS{model}=="Amazon Elastic Block Store", PROGRAM="/usr/local/bin/ebs-nvme-mapping /dev/%k", SYMLINK+="%c"

[fedora@mesos-master01fu2 ~]$ cat /usr/local/bin/ebs-nvme-mapping
#!/bin/bash

vol=$(/usr/sbin/nvme id-ctrl --raw-binary "$1" | cut -c3073-3104 | tr -s ' ' | sed 's/ $//g')
vol=${vol#/dev/}
if [[ -n "$vol" ]]; then
  echo ${vol/xvd/sd} ${vol/sd/xvd}
fi

[fedora@mesos-master01fu2 ~]$ grep nvme /usr/bin/rexray
Binary file /usr/bin/rexray matches
[fedora@mesos-master01fu2 ~]$ rexray version
REX-Ray
-------
Binary: /usr/bin/rexray
SemVer: 0.11.2+20+dirty
OsArch: Linux-x86_64
Commit: d5cf3ed1aac75a4595760fc0830beea5886588f0
Formed: Fri, 18 May 2018 17:55:35 UTC
```

jippi:
@JustinVenus okay, looks identical to my instances, couldn't get it to work - i think the PR I ended up with (#1252) is quite a bit simpler, and way less blast radius - feel free to take it for a spin if you have time :)

akutz:
Closed for #1252. Thank you @JustinVenus, you paved the way! NVMe support is finally merged.

jippi:
@romankuzmik no, the team has been unreasonable slow at releasing this :(

a-nldisr:
@akutz Any plans in releasing this to latest artifacts?

akutz:
Hi @a-nldisr,

This was merged to master, it should be available for installation as `unstable` and on the Docker repo as the `edge` tag. As for a GA release, I defer to @clintkitson. Thank you.

PR #1240, opened 2018-06-19
Resolve device names for attached Cinder volumes

rionpy:
As highlighted in https://github.com/rexray/rexray/issues/1183, Cinder suggests a device name for a volume, which is not always used by the hypervisor. This issue is reliably reproducible with force detach:
1. Mount Cinder volume on instance
2. Force detach volume from outside instance
3. Re-mount volume on instance

Doing this leads to Cinder always offering the same device name, while the hypervisor will pick something else (possibly due to abruptly losing said device).

This patch aims to resolve the issue by resolving the device names by matching the volume IDs to the `virtio-XXXXXXXX-XXXX-XXXX-X`-symlinks created under /dev/disk/by-id (adapted from https://github.com/ederst/rexray/commit/0f57fc77402f8e04b577a2feb8da294cb930981d)

### DIFF
--- a/libstorage/drivers/storage/cinder/storage/cinder_storage.go
+++ b/libstorage/drivers/storage/cinder/storage/cinder_storage.go
@@ -7,6 +7,9 @@ import (
 	"io/ioutil"
 	"net/http"
 	"time"
+	"fmt"
+	"path/filepath"
+	"os"
 
 	gofig "github.com/akutz/gofig/types"
 	"github.com/akutz/goof"
@@ -301,7 +304,7 @@ func translateVolumeV1(
 			libstorageAttachment := &types.VolumeAttachment{
 				VolumeID:   attachment["volume_id"].(string),
 				InstanceID: &types.InstanceID{ID: attachment["server_id"].(string), Driver: cinder.Name},
-				DeviceName: attachment["device"].(string),
+				DeviceName: resolveDeviceName(attachment["volume_id"].(string), attachment["device"].(string)),
 				Status:     "",
 			}
 			attachments = append(attachments, libstorageAttachment)
@@ -330,7 +333,7 @@ func translateVolume(
 			libstorageAttachment := &types.VolumeAttachment{
 				VolumeID:   attachment.VolumeID,
 				InstanceID: &types.InstanceID{ID: attachment.ServerID, Driver: cinder.Name},
-				DeviceName: attachment.Device,
+				DeviceName: resolveDeviceName(attachment.VolumeID, attachment.Device),
 				Status:     "",
 			}
 			attachments = append(attachments, libstorageAttachment)
@@ -672,7 +675,26 @@ func (d *driver) VolumeAttach(
 			fields, "error waiting for volume to attach", err)
 	}
 
-	return volume, volumeAttach.Device, nil
+	return volume, resolveDeviceName(volumeID, volumeAttach.Device), nil
+}
+
+func getDeviceLink(
+	volumeID string) string {
+
+	return fmt.Sprintf("/dev/disk/by-id/virtio-%s", volumeID[:20])
+}
+
+func resolveDeviceName(
+	volumeID string,
+	deviceName string) string {
+
+	attachedDeviceLink := getDeviceLink(volumeID)
+	attachedDeviceName, err := filepath.EvalSymlinks(attachedDeviceLink)
+	if err != nil {
+		return deviceName
+	}
+
+	return attachedDeviceName
 }
 
 func (d *driver) VolumeDetach(
@@ -766,6 +788,9 @@ func (d *driver) waitVolumeAttachStatus(
 	if volumeID == "" {
 		return nil, goof.WithFields(fields, "volumeId is required")
 	}
+
+	attachedDeviceLink := getDeviceLink(volumeID)
+
 	begin := time.Now()
 	for time.Now().Sub(begin) < timeout {
 		volume, err := d.VolumeInspect(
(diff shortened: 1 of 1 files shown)


bartmatthaei:
When can we expect this pull request to come through? We're really waiting for this fix :-)

libstorage/drivers/storage/cinder/storage/cinder_storage.go
```diff
@@ -672,7 +675,26 @@ func (d *driver) VolumeAttach(
 			fields, "error waiting for volume to attach", err)
 	}
 
-	return volume, volumeAttach.Device, nil
+	return volume, resolveDeviceName(volumeID, volumeAttach.Device), nil
+}
+
+func getDeviceLink(
```
codenrhoden: Resolving devices that are on the node by peaking in `/dev` cannot be done in this code, as it may not be running on the same node. REX-Ray supports a client/server mode of operation, where a single REX-Ray instance running in server mode is configured with the Cinder API credentials, but all client nodes are just talking to the master REX-Ray instance.

Any code that is specific to a single node/instance needs to be in the executor, not here.

The way this should be handled is in the `LocalDevices` function in the [executor](https://github.com/rexray/rexray/blob/d14c3e3073be0b0955648b7de8a0a882a87977d7/libstorage/drivers/storage/cinder/executor/cinder_executor.go#L203-L224). This function returns a map of attached devices, with their actual path to their volume ID. Right now, this function is returning all attached devices, and an empty volume ID. This would need to modified to extract the volume ID from  your `virtio` path used below...

libstorage/drivers/storage/cinder/storage/cinder_storage.go
```diff
@@ -672,7 +675,26 @@ func (d *driver) VolumeAttach(
...
+	volumeID string) string {
+
+	return fmt.Sprintf("/dev/disk/by-id/virtio-%s", volumeID[:20])
```
codenrhoden: The bit about the executor aside, is this universal across OpenStack/Cinder? I think one of the reasons things are the way they are now is because not everything is using virtio, correct? What if this was on a Xen hypervisor? Would this path be the same?

I recognize things are broken right now, and we should target the most common usecase, which is probably KVM with virtio. But if that's not universal, this needs to be a configurable behavior.

libstorage/drivers/storage/cinder/storage/cinder_storage.go
```diff
@@ -330,7 +333,7 @@ func translateVolume(
 			libstorageAttachment := &types.VolumeAttachment{
 				VolumeID:   attachment.VolumeID,
 				InstanceID: &types.InstanceID{ID: attachment.ServerID, Driver: cinder.Name},
-				DeviceName: attachment.Device,
+				DeviceName: resolveDeviceName(attachment.VolumeID, attachment.Device),
```
codenrhoden: This instead needs to match the volume ID to those found in the LocalDevices map, which right now is empty, and pull the device name from that. This code may be running on a different node than where the volume is attached.

Each driver has to do this a bit differently, but here is an example from [EBS](https://github.com/rexray/rexray/blob/d14c3e3073be0b0955648b7de8a0a882a87977d7/libstorage/drivers/storage/ebs/storage/ebs_storage.go#L953-L971) and [ScaleIO](https://github.com/rexray/rexray/blob/d14c3e3073be0b0955648b7de8a0a882a87977d7/libstorage/drivers/storage/scaleio/storage/scaleio_storage.go#L214-L230)

ederst:
Hi there @codenrhoden,
So to recap, just so that I understand this: The current proposed change won't work for all deployments of rex ray b/c stuff can run on different nodes and not everyone is using KVM+virtio, of course (this was probably also the reason the earlier attempts were rejected as well I guess).

To fix this the following is needed to be done (feel free to correct me if I'm wrong):
* move virtio device resolution to executor (cinder_executor.go -> LocalDevices function), since this runs on the machine where the device will be mounted
* get device information in cinder_storage.go from this LocalDevices generated map
* make KVM+virtio stuff configurable (or use normal cinder device resolution as fallback when no virtio device is found as suggested by @bartmatthaei)

I could try to give this a go, but since my current workload is at an all-time high (at least it feels this way), it probably will take at least a month.

libstorage/drivers/storage/cinder/storage/cinder_storage.go
kumy: `virtio-%s` this won't work in all cases, as for example in my case, devices are named as:

```
/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_2afa0091-d1c5-46e2-9
/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_426c308d-a355-4cca-a
/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_4dfda350-9cd9-4901-8
/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_52645ecc-2ac0-4534-a
/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_5569888b-5d88-4a57-9
/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_698a3ded-990a-420b-8
```

(Openstack from OVH)

hervenicol: ICYMI, for OVH/Openstack I hacked a variant and made the plugin available here: https://hub.docker.com/r/hervenicol/rexray.

hervenicol: ...it appears the hack is not enough, we still have a lot of issues: wrong volume mounted, volumes created although already existing, unable to mount volumes...
So I decided to go to a simpler, cinder-dedicated plugin: https://github.com/hervenicol/docker-plugin-cinder

It means a lot less code, easier to understand / debug / evolve.
It seems rexray was overkill for my use case. And on top of that, too complex for me to maintain / fix it by myself.

PR #1243, opened 2018-07-09
update naming convension for GCP disk

boylee1111:
This patch fixes a bug regarding GCP naming convention.


akutz:
Hi @boylee1111,

Did you see @clintkitson's note above?

> Is this change relevant globally across all GCP instances? Can anyone else validate this change?

PR #1256, opened 2018-08-15
Add AUTHORS file

codenrhoden:
@clintkitson here is what we discussed with @jonasrosland.

Is this what you were picturing?

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,0 +1,5 @@
+# REX-Ray Maintainers
+
+* [Andrew Kutz, @akutz](https://github.com/akutz)
+* [Clinton Kitson, @clintkitson](https://github.com/clintkitson)
+* [Travis Rhoden, @codenrhoden](https://github.com/codenrhoden)


akutz:
So I know we use md for a lot, but for parsing, should this be plain text? Or is this a GitHub-ism?

-- 
-a

codenrhoden:
@akutz I have no strong preference here. I just wanted there to be clickable links for names.

I don't know that there are any plans for parsing this, just for information.

@jonasrosland pointed me to [this](https://github.com/discourse/discourse/blob/master/docs/AUTHORS.md) and [this](https://github.com/facebookincubator/dhcplb/blob/master/AUTHORS) as examples. So, one MD and one non-MD.

codenrhoden:
@akutz pushed a new version after reviewing with Jonas and Clint today. Changed to "REX-Ray Maintainers" to indicate roles. They were good with the .md version - no current plans for parsing this file.

jonasrosland:
LGTM!

PR #1265, opened 2018-09-23
add `S3FS_CONVERTUNDERSCORES` env into s3 conf.json

chaiebaymen:
the purpose is to take choice to convert underscore between my stack name and the service. because as you know aws S3 don't support `_` in buckets names.
please i wish that you will making some stuff for that.
i'm waiting for your feedback and thanks for your attention.

### DIFF
--- a/.docker/plugins/s3fs/config.json
+++ b/.docker/plugins/s3fs/config.json
@@ -130,7 +130,15 @@
             "value"
           ],
           "Value": "false"
-        }
+        },
+        {
+          "Description": "",
+          "Name": "S3FS_CONVERTUNDERSCORES",
+          "Settable": [
+            "value"
+          ],
+          "Value": "false"
+        }            
       ],
       "Interface": {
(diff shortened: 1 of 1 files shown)


clintkitson:
Hello @chaiebaymen,

Does this fix work? I don't see this as an actual parameter. It looks like someone needs to enable this as a capability in the driver to be able to expose it to the managed plugin.

chaiebaymen:
Hello @klintkistson,
Firstly, thanks for your feedback, exactly i want to expose this alias to be able to integrate my docker volume as s3 bucket into my docker stack. Actually, once i deloy the stack ( ‘docker stack deploy -c stack-file.yml stacknam’) a volume will be mounted as s3 bucket like ´stackname_volumename’ and eventually it’s not acceptable to create bucket with underscored name. My purpose is to make an option for the plugin in order to activate the conversion of the underscore to a dash for exemple.

rimoore:
I am having the same issue.

Issue #1228, opened 2018-05-24
Can't mount EFS volumes in other subnets


jwslok:
# Summary
I would like to mount existing EFS volumes with existing mount targets in a data subnet to an EC2 instance in another subnet (same AZ).

# Bug Reports
The current version of REX-Ray looks for mount targets in the subnet of the EC2 instance. However, our setup uses separate data subnets for storage resources. We would like to mount a filesystem "/foobar" with an existing mount target in AZ "eu-west-1a" and subnet "data" to an EC2 instance in the same AZ with subnet "private". Currently this is not possible because REX-Ray looks for mount targets in the EC2 instances subnet, which are not available.

## Version
```shell
rexray/efs:edge and rexray/efs:latest
```

## Expected Behavior
When starting a Docker container that will use existing volume '/foobar' with existing mount targets in the same AZ, I expect REX-Ray to discover the mount target and successfully create the volume mount. 

## Actual Behavior
When the existing volume does not have a mount target in the same subnet we're running in, it will produce an error:

```
level=error msg=\"docker-legacy: Mount: foobar: failed: resource not found\" host=\"unix:///var/run/rexray/670574597.sock\" integrationDriver=linux osDriver=linux service=efs storageDriver=libstorage
```

(Note: manually mounting the same volume works fine)

## Steps To Reproduce

1. Create an EC2 instance in AZ "eu-west-1a" and subnet "data"
2. Create an EFS volume "/foobar"
3. Create a mount target for "/foobar" in AZ "eu-west-1a" and  subnet "private"
4. Run REX-Ray Docker plugin
5. Run a Docker container and mount /foobar
```
docker run -d --volume-driver=rexray/efs -v foobar:/foobar nginx
```

## Configuration Files

```shell
# Install REX-Ray EFS plugin
docker plugin install --grant-all-permissions rexray/efs \
EFS_ACCESSKEY=${IAM_ACCESS} \
EFS_SECRETKEY=${IAM_SECRET} \
EFS_REGION=eu-west-1 \
LINUX_VOLUME_FILEMODE=0777
```

## Logs
```
/run/torcx/bin/docker: Error response from daemon: error while mounting volume '/var/lib/docker/plugins/<ID>/rootfs': VolumeDriver.Mount: docker-legacy: Mount: foobar: failed: resource not found.
```

### Service Log
```
[...]

PR #1268, opened 2018-09-29
WiP - Look at availability zones when searching for EFS mount targets

SijmenHuizenga:
This pull-request contains changes to fix the problem described in #1228. 
Before this commit the EFS driver would select a mount target based on the subnet it is in. All mount targets in subnets not equal to the subnet of the host would be ignored. This is incorrect behaviour as subnets do not restrict access to a mount target. Now the mount target selection is implemented by looking at the availability zone of the mount target and host as these must be equal for the host to reach the mount target.

*This pull-request is not ready to be merged. The following things must be done first:*

- [ ] Testing. I tested these changes a bit but not enough for release. Please test in some different situations.
- [ ] EC2 permissions documentation and testing: It is not possible to request the availability zone of a mount target directly through the aws efs api. A workaround is to request the availability zone of the subnet of the mount target as a mount target always lives in 1 availability zone and has 1 subnet. To do this, rexray needs to access the aws ec2 api. I implemented this but have not tested if any new authorization parameters are required. This should be investigated and documented.
- [ ] Code Review. This was my first time working on such a big go project so please revew the code before taking this into production.

Because of reasons, i won't be able to finish the work that i started here :disappointed: . I hope this helps someone in the right direction of a permanent fix :smile:

### DIFF
--- a/libstorage/drivers/storage/efs/efs.go
+++ b/libstorage/drivers/storage/efs/efs.go
@@ -37,6 +37,10 @@ const (
 	// group value from the InstanceID Field map.
 	InstanceIDFieldSecurityGroups = "securityGroups"
 
+	// InstanceIDFieldSubnetID is the key of the subnet where the instance lives from the
+	// InstanceID Field map
+	InstanceIDFieldSubnetID = "subnet"
+
 	// AccessKey is a key constant.
 	AccessKey = "accessKey"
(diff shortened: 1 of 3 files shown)


SijmenHuizenga:
Closing this in favour of #1288

PR #1288, opened 2018-12-19
1228 efs avz select

mtsr:
Since we're running into the same issue in #1228 I'd like to pick up Sijmen Huizenga's work from #1268.

I've fixed the hound comments. Could a maintainer take a look if something else needs to be done?

@SijmenHuizenga Could you sign the CLA?

### DIFF
--- a/libstorage/drivers/storage/efs/efs.go
+++ b/libstorage/drivers/storage/efs/efs.go
@@ -37,6 +37,10 @@ const (
 	// group value from the InstanceID Field map.
 	InstanceIDFieldSecurityGroups = "securityGroups"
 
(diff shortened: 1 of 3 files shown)


clintkitson:
We are looking for the community to validate the testing in AWS for this patch. @SijmenHuizenga Any feedback for this continuation of your work?

SijmenHuizenga:
@clintkitson Sorry for the embarrassing late response. I won't be working on this anymore.

lhenry7:
@clintkitson  Hi I am also trying to put my EFS mount points in a different subnet than the Docker host, and I need this change in order for it to work. Is this change going to be finished, or should I look for a different solution?

Issue #1212, opened 2018-05-01
Rexray VolumeDriver.Mount: ControllerPublishVolume failed w/ DOBS


DreadfulDeveloper:
# Summary
When deploying a stack using a compose file, the service that has a rexray volume attached never comes online (stays replicated 0/1).

# Bug Reports
This section is for issues that relate to discovered problems or bugs.
When running ```docker service ps --no-trunc [id]``` this is listed as the failing error for the service.

```
VolumeDriver.Mount: ControllerPublishVolume failed: 0: ControllerPublishVolume failed: 0: rpc error: code = Unknown desc = unable to find attached vol in local devices"
VolumeDriver.Mount: ControllerPublishVolume failed: 0: ControllerPublishVolume failed: 0: rpc error: code = Unknown desc = unable to find attached vol in local devices"
VolumeDriver.Mount: NodePublishVolume failed: 6: device not found: timed out"
VolumeDriver.Mount: ControllerPublishVolume failed: 0: ControllerPublishVolume failed: 0: rpc error: code = Unknown desc = unable to find attached vol in local devices"
```

## Version

```shell
$ rexray version
REX-Ray
-------
Binary: /usr/bin/rexray
Flavor: client+agent+controller
SemVer: 0.11.2
OsArch: Linux-x86_64
Commit: fc8bfbd2d02c2690fc3a755a9560dd12c88e0852
Formed: Sat, 24 Feb 2018 23:51:22 UTC
```

## Expected Behavior
When deploying a service with a rexray volume, the volume is created and attached.

## Actual Behavior
The service with the rexray volume never comes online, because .

## Steps To Reproduce
1. Install rexray as service
    * curl -sSL https://rexray.io/install | sh
2. Add config.yml (listed below)
3. Restart rexray service
    * ```rexray service restart```
4. deploy stack via compose file with rexray volume (example compose below)
    * ```docker stack deploy -c docker-compose.yml doesnt-work```
5. get service id from docker
    * ```docker service ls```
5. watch service for failures
    * ```watch docker service ps --no-trunc```

### Rexray config
```yaml
rexray:
  logLevel: debug
libstorage:
  logging:
    level: debug
  service: dobs
  integration:
    volume:
      operations:
        create:
          default:
            size: 1
        mount:
          preempt: true
          rootPath: /data
dobs:
  token: 123456
  region: nyc1
```
### Example compose - deployed as stack

```yaml
version: '3.4'
services:
  wordpress:
    image: wp
    restart: always
[...]

PR #1359, opened 2021-02-04
DOBS: retry on volume action returning 404

Galbar:
I think this may be able to solve #1212 

As I say in my comment in that issue: I believe that this happens because DO doesn't have time to sync behind the scene between returning the volume attach action id and the polling that rexray does for the action to complete (This is a theory, I have not tested it).

I tried to run it locally but I can't get go to work :(

I wanted to share the idea to get feedback on the approach.

I just noticed there is another PR that implements the same fix: https://github.com/rexray/rexray/pull/1344


Galbar:
Closing in favour of https://github.com/rexray/rexray/pull/1344

Issue #1255, opened 2018-08-15
rexray/ebs plugin throwing error /run/docker/plugins/03c113e38e92e1482675c82db3554c88e5ebb50cc42d1ecdb0a54f5a3e9035b8/rexray.sock: connect: no such file or directory


Amir345:
rexray/ebs plugin throwing error /run/docker/plugins/03c113e38e92e1482675c82db3554c88e5ebb50cc42d1ecdb0a54f5a3e9035b8/rexray.sock: connect: no such file or directory

I am trying to install the rexray/ebs plugin using the official documentation but it is not working as expected

however when installing s3fs plugin it works fine

Plugin "rexray/ebs" is requesting the following privileges:
 - network: [host]
 - mount: [/dev]
 - allow-all-devices: [true]
 - capabilities: [CAP_SYS_ADMIN]
Do you grant the above permissions? [y/N] y
latest: Pulling from rexray/ebs
d726edfb6c8d: Download complete
Digest: sha256:b5b0680b6dd53a4c61bac21942738a69edfe3d6a6516acc13045d257889c9213
Status: Downloaded newer image for rexray/ebs:latest
Error response from daemon: dial unix /run/docker/plugins/93e0f96ed757cd900d30232ef9c321c3956fe74bc9c56119a43ec1642869fd9b/rexray.sock: connect: no such file or directory

clintkitson:
Have you checked the Docker daemon logs @Amir345?

Amir345:
level=error msg="error: service startup failed: agent: mod init failed: error initializing instance ID cache" plugin=93e0f96ed757cd900d30232ef9c321c3956fe74bc9c56119a43ec1642869fd9b

level=error msg="Sending SIGTERM to plugin failed with error: process 93e0f96ed757cd900d30232ef9c321c3956fe74bc9c56119a43ec1642869fd9b not found: not found"

level=error msg="Handler for POST /v1.38/plugins/rexray/ebs/enable returned error: dial unix /run/docker/plugins/93e0f96ed757cd900d30232ef9c321c3956fe74bc9c56119a43ec1642869fd9b/rexray.sock: connect: no such file or directory"

Amir345:
@clintkitson here is the docker daemon logs which i found.

clintkitson:
Have you tried starting the plugin with `REXRAY_DEBUG=true` as an env variable and checking the logs? It is very likely that your problem is authentication based.

qianzhangxa:
[...]

PR #1372, opened 2022-11-25
fix metadata (IMDSv2) in EBS driver

### LINKED ISSUE #1371: rexray/ebs plugin is not working with EC2 metadata version 2
# Summary

We're making our EC2 instances to work with instance metadata Version 2 (only) for some security reasons. After we enabled the metadata version 2 on EC2s, rexray/ebs plugin is not working on those instances. Does rexray/ebs plugin has support for EC2 metadata version 2 (only) ?

# Bug Reports
rexray/ebs plugin is not working with EC2 metadata version 2

## Actual Behavior
`Error response from daemon: dial unix /run/docker/plugins/185d790355bc4f0515af8925fa4aba54196de9f57d5a17686a975d0fc2f36856/rexray.sock: connect: no such file or directory`

nbryant42:
Fixes #1371 which is important because IMDSv2 is the new security best practice and is starting to become required for compliance with SOC2 etc.

Also migrates from the deprecated and broken `dep` vendoring tool to `go mod`. 

I have gone to great lengths to ensure that all package versions remain the same as they were under `dep`. All versions are as they used to be, except for `aws-sdk-go`, which I obviously had to upgrade slightly. `go mod` ignores transitive test dependencies, so there are a lot of deletions under `vendor/`, but these are only deletions of packages that are unnecessary for the build, with the obvious exception of `awk-sdk-go` and its dependencies.

Note that `go.mod` contains a `replace` statement to force the `x/sys` version to downgrade to the same version that was used under `dep`, to prevent `go mod tidy` from upgrading it. I don't know why `go mod tidy` wants to upgrade it, presumably some quirk of the minimum version selection algorithm when importing repos that lack a `go.mod`. This is the only package where I had to resort to any such shenanigans (the other `replace` statements were migrated from the previous setup.) It's probably pretty low risk to allow this package to upgrade, but I wanted to keep the versions completely consistent unless the maintainers of this repo would prefer otherwise.


rixwan-sharif (approved):

rmccarthy-ellevation:
Any chance we can get this merged?

romed:
This probably also fixes #1255

